PR #4, opened 2019-09-08
Support using with std-enabled serde

dtolnay:
Previously, using Postcard in a project that also uses std-enabled Serde did not work.

```toml
[dependencies]
serde = "1.0" # only works if we add default-features=false
postcard = "0.4"
```

```console
error[E0277]: the trait bound `error::Error: std::error::Error` is not satisfied
  --> postcard-0.4.0/src/error.rs:71:6
   |
71 | impl serde::ser::Error for Error {
   |      ^^^^^^^^^^^^^^^^^ the trait `std::error::Error` is not implemented for `error::Error`
```

This PR makes Postcard compile whether or not Serde std is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ version = "0.5"
 features = ["serde"]
 
 [dependencies.serde]
-version = "1.0"
+version = "1.0.100"
 default-features = false
 features = ["derive"]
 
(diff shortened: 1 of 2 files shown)


jamesmunns:
This was released as postcard 0.4.1: https://crates.io/crates/postcard/0.4.1

PR #6, opened 2019-10-28
Improvements to serialisation with `StdVec` for `use-std` feature

spacekookie:
This PR does two things:

1. Make the inner type of `StdVec` pub, so that it can be created
   manually outside of `postcard`.
2. Add `to_stdvec` and `to_stdvec_cobs` utility functions, to make it
   easier to directly serialise to these flavours

I also added doc tests, one of which fails which is a bit
strange. It's basically the same example as the `to_vec_cobs` doc
test. And I don't assume that the `std::vec::Vec<u8>` would handle
bytes drastically different from `heapless::Vec`...right? Anyway,
maybe you have some ideas there.


jamesmunns:
This has been released on crates.io as v0.4.2. CC @spacekookie and @hannobraun

PR #9, opened 2019-12-05
Fix tuple deserialize

### LINKED ISSUE #8: Tuple struct deserialization bug
I've encountered a bug when attempting to deserialize a tuple struct. The struct I'm attempting to deserialize is:
```
#[derive(Serialize, Deserialize)]
pub struct DualTupleStruct(u8, u16)
```
However, deserializaiton fails with:
```
Err(SerdeDeCustom)
```

This structure is different than that which is covered in the test cases:
```
#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
pub struct TupleStruct((u8, u16))
```

I've tracked down the issue to deserialize_tuple_struct() in the Deserializer. It attempts to deserialize tuple structs as a sequence, when they should be deserialized as a tuple.

yungkneez:
Fixes #8

### DIFF
--- a/src/de/deserializer.rs
+++ b/src/de/deserializer.rs
@@ -301,13 +301,13 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     fn deserialize_tuple_struct<V>(
         self,
         _name: &'static str,
-        _len: usize,
+        len: usize,
         visitor: V,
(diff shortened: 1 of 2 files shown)


jamesmunns:
Releases as [v0.4.3](https://crates.io/crates/postcard/0.4.3).

Issue #16, opened 2020-04-08
Serialize/deserialize with footer?


korken89:
What is the current recommended way to serialize/deserialize with a footer?
With this I mean, in my usecase, I want to append a CRC to the end of the buffer and each byte being serialized should also go through the CRC calculation.
Currently if I serialize with a flavor I cannot append the CRC manually as it can for example have been encoded with COBS.

Let's say I want something like this:

```rust
struct Message {
    // ...
}
pub fn serialize_into_buffer(msg: Message, buf: &mut [u8]) -> postcard::Result<usize> {
 
    // How to add footer? A "add footer flavor" which takes a closure?

    let res = serialize_with_flavor(&msg, Cobs::try_new(Slice::new(buf))?)?;
    Ok(res.len())
}
```

korken89:
The issue I see with a "footer flavor" is that there is no way to push the footer before COBS finishes.
One could extend the COBS flavor to have an optional "generate footer" that is `None` in the normal case, and if there is something to generate a footer it will be pushed before the finalization of COBS, but then one would have to add so the deserialization returns an optional footer.

korken89:
An alternative is to implement Hash for it as such:

```rust
#[derive(Hash, Serialize, Deserialize)]
pub enum Message {
    /// ...
}

#[derive(Serialize, Deserialize)]
struct Packet {
    msg: Message,
    crc: u16,
}
```

And fill the CRC with a suitably implemented hash.

bugadani:
I'm necromancing this issue because I am also looking for similar functionality. I'm not really familiar with Postcard at this time, so this might not be a good idea: My approach would be a CRC flavour that wraps COBS. Dropping the flavour object could/would write the computed CRC with its "last breath" to the COBS flavour - although a more elegant solution would be to expose the end of the message to the flavours.

jamesmunns:
I'm going to close this, as I think this is possible with flavors, which can be stacked. For example, you could have a flavor like `Cobs<Crc<Slice>>`. That being said, mixing cobs and crc is [a complicated question](https://bsky.app/profile/jamesmunns.com/post/3lgnijtyebc2k).

[`Flavor::finalize()`](https://docs.rs/postcard/latest/postcard/ser_flavors/trait.Flavor.html#tymethod.finalize) can return arbitrary data (or perform arbitrary calculations) of your choice when finalizing the serialization.

PR #17, opened 2020-04-08
COBS with footer PoC

korken89:
I made an PoC for https://github.com/jamesmunns/postcard/issues/16, feedback is very welcome!
The deserailization is a bit hacky right now, but it shows the concept, if no footer is requested from the new method the code is exactly as it was before without bloat.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,7 +132,7 @@
 //!
 //! let data: &[u8] = &[0x01, 0x00, 0x20, 0x30];
 //! let buffer = &mut [0u8; 32];
-//! let res = serialize_with_flavor::<[u8], Cobs<Slice>, &mut [u8]>(
+//! let res = serialize_with_flavor(
 //!     data,
 //!     Cobs::try_new(Slice::new(buffer)).unwrap(),
 //! ).unwrap();
@@ -167,9 +167,7 @@ mod varint;
 pub use de::deserializer::Deserializer;
 pub use de::{from_bytes, from_bytes_cobs, take_from_bytes, take_from_bytes_cobs};
 pub use error::{Error, Result};
-pub use ser::{
-    flavors, serialize_with_flavor, serializer::Serializer, to_slice, to_slice_cobs,
-};
+pub use ser::{flavors, serialize_with_flavor, serializer::Serializer, to_slice, to_slice_cobs};
 
 #[cfg(feature = "heapless")]
 pub use ser::{to_vec, to_vec_cobs};
--- a/src/ser/flavors.rs
+++ b/src/ser/flavors.rs
@@ -63,7 +63,7 @@
 //!
 //! let data: &[u8] = &[0x01, 0x00, 0x20, 0x30];
 //! let buffer = &mut [0u8; 32];
-//! let res = serialize_with_flavor::<[u8], Cobs<Slice>, &mut [u8]>(
+//! let res = serialize_with_flavor(
 //!     data,
 //!     Cobs::try_new(Slice::new(buffer)).unwrap(),
 //! ).unwrap();
@@ -183,10 +183,10 @@ impl<'a> IndexMut<usize> for Slice<'a> {
 
 #[cfg(feature = "heapless")]
 mod heapless_vec {
-    use heapless::{ArrayLength, Vec};
-    use super::SerFlavor;
     use super::Index;
     use super::IndexMut;
+    use super::SerFlavor;
+    use heapless::{ArrayLength, Vec};
 
     ////////////////////////////////////////
     // HVec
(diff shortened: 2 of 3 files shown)


korken89:
This should probably not be accepted, it's just a PoC. :)

jamesmunns:
So, I need to take a closer look, but I think the "right" way to do this would be just with stacking flavors. You can think if it like the OSI model, every flavor is entirely wrapped by the next layer down, and you can add whatever you want (or modify parts in the middle) for each flavor. So if you wanted to append a CRC32 to cobs encoded data (and put it into a slice), you would do:

`CRC32Flavor<CobsFlavor<SliceFlavor<T>>>`

If you wanted to have the CRC32 *inside* of the COBS encoding, you would do:

`CobsFlavor<CRC32Flavor<SliceFlavor<T>>>`

Let me know if I misunderstood, I'll take a closer look at this after I process the `Mu` PR!

korken89:
Thanks for the feedback!
I had a look at the `CobsFlavor<CRC32Flavor<SliceFlavor<T>>>`, as it would do what I want.
The issue is that the current `CobsFlavor` would not finish the `CRC32Flavor`, letting its bytes be written, before `CobsFlavor` finishes itself, see this line:
https://github.com/jamesmunns/postcard/blob/46ff6514fd93f82f906edf5b6eb4f4c00f3ef9f8/src/ser/flavors.rs#L402

What I think is missing in the `SerFlavor` trait is an optional `finalize` method, as this flavor needs to finalize and push its bytes.

Or I might have misunderstood your idea here?

korken89:
Hi, @jamesmunns 

Do you have feedback here?
I'm still looking into how to have "footers" generated with postcard, mainly for checksum generation.

What do you think about the `finalize` extension to the trait?

jamesmunns:
Hey @korken89, I'll look at this after Oxidize, feel free to reping me then!

MathiasKoch:
Is this completely stale?

PING @jamesmunns @korken89

korken89:
This was mostly a way to hack it in, so its not something that should be used.
I have added earlier what is needed from my PoV for this concept to be possible.

For now I am doing this as a workaround (part of the code):

```rust
/// Serialize a message into a buffer with a provided CRC16-CCITT hasher.
pub fn serialize_with_crc16<H: Hasher>(
    hasher: &mut H,
    msg: &Message,
    buf: &mut [u8],
) -> Result<usize> {
    let len = serialize_with_flavor(msg, Slice::new(buf))?.len();

    // Apply CRC
    hasher.write(&buf[0..len]);
    let crc = (hasher.finish() as u16).to_le_bytes();
    buf[len] = crc[0];
    buf[len + 1] = crc[1];

    // Apply COBS encoding
    let len = cobs::encode(&buf[0..len + 2], buf);
    buf[len] = 0; // Add the ending 0 (cobs crate does not handle that)

    Ok(len + 1)
}
```

blaind:
I did it like:

```
use crc32fast::Hasher;

pub struct Crc32Flavor<B> where B: SerFlavor {
    flav: B,
    hasher: Hasher,
}

impl<B> Crc32Flavor<B> where B: SerFlavor {
    pub fn new(bee: B) -> Self {
        let hasher = Hasher::new();
        Self { flav: bee, hasher }
    }
}

impl<'a, B> SerFlavor for Crc32Flavor<B> where B: SerFlavor {
    type Output = <B as SerFlavor>::Output;
  
    fn try_push(&mut self, data: u8) -> Result<(), ()> {
        self.hasher.update(&[data]);
        self.flav.try_push(data)
    }
  
    fn release(mut self) -> Result<Self::Output, ()> {
        let hash = self.hasher.finalize();
        for byte in hash.to_le_bytes().iter {
            self.flav.try_push(*byte)?;
        }
    
        self.flav.release();
    }
}
```

and then using Crc32Flavor as the "outermost" flavor.

korken89:
Thanks for the example @blaind !
I will have to take a look and compare with my old notes on the issues I saw, it looks like I was overthinking the issue :)

korken89:
One question @blaind , when you deserialize - how do you perform the extraction of the CRC?

blaind:
Btw I noticed that crc32fast Hasher results to really large binary - had to switch to another method, but also that hasher takes u32:s in instead of u8 as in above example - that made the solution a bit more complex (for now), but I guess it can be abstracted away to the hasher implementation, if needed.

Anyway, about the extraction:

1. Decode cobs (using `cobs` crate)
2. Get output slice that's basically [u8]: [data..., c4, c3, c2, c1]
3. Convert c4-c1 slice to u32 (`u32::from_le_bytes`). Basically I'll just make sure that slice.len() is at least 4, and take the last 4 bytes out
4. Calculate crc32 for data...
5. Compare crc32 & u32(c4-c1)

jamesmunns:
Closing this as out of date, probably a dupe of #117

PR #22, opened 2020-06-05
Add missing size prefix for serialize_bytes

### LINKED ISSUE #21: Serializer::serialize_bytes does not serialize the size prefix
The `Serializer::serialize_bytes` should include the number of bytes serialized so that `Deserializer::deserialize_bytes` can correctly read the number of bytes back when deserializing, but the current `serialize_bytes` implementation omits this. This breaks serialization for types that use `serialize_bytes` and `deserialize_bytes`, such as the `Bytes` and `BytesMut` types from the `bytes` crate or `CStr` and `CString`.

#### Cargo.toml

```toml
[package]
name = "serialize_bytes_test"
version = "0.1.0"
edition = "2018"

[dependencies]
bytes = { version = "0.5.4", features = ["serde"] }
heapless = "0.5.5"
postcard = "0.5.0"
```

#### src/main.rs

```rust
use bytes::BytesMut;
use heapless::{consts::U16, Vec};
use postcard::{from_bytes, to_vec};

fn main() {
    let x = BytesMut::from(&[1u8, 2, 3, 4][..]);
    let serialized: Vec<u8, U16> = to_vec(&x).unwrap();
    let deserialized: BytesMut = from_bytes(&serialized).unwrap();
    assert_eq!(deserialized, x);
}
````

okready:
Serializer::serialize_bytes does not write the buffer size when
serializing, preventing deserialization from working properly when
serializing types that make use of this method (e.g. Bytes and BytesMut
from the bytes crate, CStr and CString from std::ffi). Simply adding the
missing size serialization breaks all size prefix serialization, as size
prefix serialization uses the existing Serializer::serialize_bytes
implementation to write the VarintUsize bytes to the output flavor.

This commit adds a try_push_varint_usize method to the SerFlavor trait
for writing out the encoded VarintUsize bytes, replacing all
VarintUsize::serialize calls with the SerFlavor method call, and adds
the same call to Serializer::serialize_bytes to ensure the size prefix
is now properly written. The VarintUsize::serialize implementation has
also been updated to serialize the encoded bytes as a tuple, mostly
retaining compatibility with the previous implementation while using
postcard's serializer, but potentially breaking compatibility with other
serializers. A deserialization test has also been added for
serialize_bytes output to mitigate possible regressions.

Fixes #21.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,6 @@ version = "0.1.5-pre"
 default-features = false
 
 [features]
-use-std = []
+use-std = ["serde/std"]
 default = ["heapless"]
 alloc = ["serde/alloc"]
--- a/src/de/deserializer.rs
+++ b/src/de/deserializer.rs
@@ -262,8 +262,6 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
-        // AJM - in serialize_bytes, we don't write the length first
-        // is this asymmetry intended?
         let sz = self.try_take_varint()?;
         let bytes: &'de [u8] = self.try_take_n(sz)?;
         visitor.visit_borrowed_bytes(bytes)
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -59,7 +59,7 @@ mod test_heapless {
     use core::fmt::Write;
     use core::ops::Deref;
     use heapless::{consts::*, String, Vec, FnvIndexMap};
-    use serde::{Deserialize, Serialize};
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
     #[test]
     fn de_u8() {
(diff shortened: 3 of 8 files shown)


jamesmunns:
Hi @okready! Thanks for taking care of this. To be clear - this doesn't change the "on the wire" representation for any values, correct? I admit I don't totally grok what the problem was/what this fixes, but that's probably my lack of understanding of this part of the serde machinery!

okready:
Yeah, I re-read my explanation and realized it's pretty bad. :) In short, the wire representation should remain the same.

To try to clarify, `Serializer::serialize_bytes` essentially allows for serializing slices of bytes without making separate individual calls to `serialize_seq` and `serialize_element`, so it needs to include the length when serializing, otherwise we break serialization of any types that use it (such as the `bytes` crate and `CStr`/`CString`). Unfortunately, you can't just serialize a `VarintUsize` at the top of the `serialize_bytes` call since `VarintUsize::serialize` itself calls `serialize_bytes`, so you also need to change how varints are written.

Simply changing the `Serialize` implementation for `VarintUsize` is pretty limiting since you're abstracted from the `Serializer` internals, so instead I added a method to `SerFlavor` to directly write the varint bytes to the output buffer. It has a default implementation, so existing code should still compile against it the same, but I believe it still warrants a minor version bump according to semver rules.

The "potentially breaking compatibility with other serializers" statement in my original commit message was because I mistakenly thought `VarintUsize` was public at the time (I saw the `pub` scope for the type but didn't look at the scope of the module itself). Since it isn't public, I've removed its `Serialize` implementation for it since it's no longer used, but I can add one back if you'd prefer. I've also updated my original commit message to try to make things clearer.

okready:
It should be possible to add a regression test that verifies the wire format for various types hasn't changed. I wouldn't be able to get to it today, put should have time tomorrow or later on this week.

okready:
I realized the `loopback` test would have failed if the wire format changed, so it looks like that should cover backwards compatibility.

I've added in a new commit though to test against `CString` within the `loopback` test to verify the `serialize_bytes`/`deserialize_byte_buf` implementations work. It's limited to the `use-std` feature and required flipping on the `serde/std` feature under that configuration (AFAIK you can't specify test-only features, but I can remove that test or put in something different if there's a reason not to enable `serde/std` when `use-std` is enabled).

jamesmunns:
Thanks @okready! I'll release this shortly.

PR #25, opened 2020-08-18
Update deserialize_char to match serialize_char

boylede:
Hi, I noticed chars where not deserializing the same as they are serializing. It looks like they are being written the same as strings, e.g. length-prepended, but the deserialize code didn't read that varint. 

Also, when adding tests to this I noticed that I couldn't get it to work with multi-byte chars. It looks like char::from_u32() doesn't strip the utf-8 encoding's upper-bits from each byte, so it has to be done manually. This would be a lot of extra code and I don't see anything in the core/standard library that just gives us a single char, so I pass this through str::from_utf8() in lieu of writing that bit manipulation here.


jamesmunns (approved):
Thanks!

PR #30, opened 2020-11-13
Add `heapless-cas` feature (enabled by default)

daniel5151:
I'm playing around with `postcard` on an custom target that doesn't support CAS. Luckily, `heapless` already includes a `cas` feature flag!

This PR adds a new enabled-by-default `heapless-cas` feature which corresponds to the `heapless/cas` feature.


jamesmunns:
Looks good, thanks!

PR #41, opened 2021-08-05
Add CobsAccumulator struct and tests

jamwaffles:
Opened as a jumping off point rather than a finished PR as there is some stuff I'm not sure about.

Questions:

- [x] Do we need a better name for `CobsAccumulator`?
- [x] Should I be feature gating anything? I don't think so but I could be wrong.
- [x] In the test, should I be trying to move any more behaviour into `CobsAccumulator` or is the API ok as is?
- [x] Should I add more tests? If so, for what test cases?
- [x] Where should I add an example? `lib.rs` (and consequently `README.md`)?

### DIFF
--- a/src/accumulator.rs
+++ b/src/accumulator.rs
@@ -35,7 +35,7 @@ impl<const N: usize> CobsAccumulator<N> {
     /// Create a new accumulator.
     pub fn new() -> Self {
         CobsAccumulator {
-            buf: unsafe { MaybeUninit::zeroed().assume_init() },
+            buf: [0; N],
             idx: 0,
         }
     }
--- a/src/accumulator.rs
+++ b/src/accumulator.rs
@@ -4,6 +4,9 @@ use serde::Deserialize;
 
 /// An accumulator used to collect chunked COBS data and deserialize it.
 ///
+/// This is often useful when you receive "parts" of the message at a time, for example when draining
+/// a serial port buffer that may not contain an entire uninterrupted message.
+///
 /// # Examples
 ///
 /// Deserialize a struct by reading chunks from a [`Read`]er.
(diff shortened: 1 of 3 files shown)


src/accumulator.rs
```diff
@@ -0,0 +1,125 @@
...
+        /// Remaining data left in the buffer after deserializing.
+        remaining: &'a [u8],
+    },
+}
+
+impl<const N: usize> CobsAccumulator<N> {
+    /// Create a new accumulator.
+    pub fn new() -> Self {
+        CobsAccumulator {
+            buf: unsafe { MaybeUninit::zeroed().assume_init() },
```
jamesmunns, line 38:
```suggestion
            buf: [0; N],
```

jamwaffles: Ooo shiny. I just pushed a `fixup!` commit. Should I keep doing that so you can rebase, or are you going to just squash and merge?

jamesmunns:
Overall looks good, with the most recent change, you can also remove the dep on MaybeUninit.

* Name is fine
* No feature gate (you didn't use heapless)
* API seems fine for now
* I can't think of any
* If you just add an example to the `///` comments on the struct and/or the functions themselves, that should be good.

src/accumulator.rs
jamesmunns: You can just discard this comment, you already did it (don't forget to remove the `use` too!

jamwaffles:
Examples added :)

src/accumulator.rs
```diff
@@ -0,0 +1,179 @@
+//! An accumulator used to collect chunked COBS data and deserialize it.
+
+use serde::Deserialize;
+
+/// An accumulator used to collect chunked COBS data and deserialize it.
+///
```
jamesmunns, line 6:
```suggestion
///
/// This is often useful when you receive "parts" of the message at a time, for example when draining
/// a serial port buffer that may not contain an entire uninterrupted message.
///
```

jamesmunns:
Thanks! Will release shortly.

PR #51, opened 2021-12-13
lib.rs: also add hint about no-std serde deps

Lotterleben:
as requested in https://github.com/jamesmunns/postcard/pull/47 :)


jamesmunns:
Thanks!

PR #54, opened 2022-01-30
Add a `MaxSize` trait and derive macro

lachlansneff:
This adds a trait `MaxSize`:

```rust
pub trait MaxSize {
    /// The maximum possible size that the serialization of this
    /// type can have, in bytes.
    const POSTCARD_MAX_SIZE: usize;
}
```

and a derive macro that implements it for structs and enums.

I think this just needs a little more documentation and then it'll be ready to merge. I would like someone else to double check my logic for computing the varint size though.

The naming needs a little more work too. I'm not super happy with the trait and constant name.

### DIFF
--- a/postcard-derive/src/lib.rs
+++ b/postcard-derive/src/lib.rs
@@ -1,0 +1,118 @@
+use proc_macro2::TokenStream;
+use quote::{quote, quote_spanned};
+use syn::{parse_macro_input, parse_quote, DeriveInput, Generics, GenericParam, Data, spanned::Spanned, Fields};
+
+/// Derive the `postcard::MaxSize` trait for a struct or enum.
+#[proc_macro_derive(MaxSize)]
+pub fn derive_max_size(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = parse_macro_input!(item as DeriveInput);
+
+    let name = input.ident;
+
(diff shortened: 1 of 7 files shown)


postcard-derive/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+            });
+
+            quote! {
+                #discriminant_size + #max
+            }
+        },
+        Data::Union(_) => unimplemented!(),
+    }
```
jamesmunns, line 72: Might want to remove this or add a better error

lachlansneff:
Here's what the error looks like when the type of a field doesn't implement `postcard::MaxSize`.

```
error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> tests/max_size.rs:62:9
   |
62 |         a: &'static str,
   |         ^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `str`
   = note: required because of the requirements on the impl of `MaxSize` for `&'static str`

error[E0277]: the trait bound `str: MaxSize` is not satisfied
  --> tests/max_size.rs:62:9
   |
62 |         a: &'static str,
   |         ^ the trait `MaxSize` is not implemented for `str`
   |
   = note: required because of the requirements on the impl of `MaxSize` for `&'static str`
```

PR #58, opened 2022-06-11
Clarify the non-self-describing nature of postcard

huntc:
Adds some more doc to the deserializer noting that postcard [is non-self-describing](https://github.com/jamesmunns/postcard/blob/main/src/de/deserializer.rs#L207). My hope is that this will help others avoid any surprises.


jamesmunns:
Thanks!

PR #60, opened 2022-06-20
Optimize Varint encoding

bugadani:
This PR optimizes the `varint_*` functions to avoid an unnecessary bit masking.

The idea behind this PR is the following:
 - Iteration stops if `(value >>= 7) == 0`, in which case `value < 128`. In this case, the `0x7F` masking isn't necessary.
 - If `value > 128` we want to set the highest bit of the extracted byte to 1. In this case, the `0x7F` masking isn't necessary, because we overwrite the masked bit anyway.

Assuming I'm not catastrophically wrong, the resulting code optimizes slightly better: https://rust.godbolt.org/z/8dz48vocx


jamesmunns:
Looks great, thanks!

PR #61, opened 2022-06-20
Fix version in changelog

bugadani:
I suspect this is a victim of a copy-paste error. `alpha.1 -> alpha.2` is written twice.


jamesmunns:
Closing this as I've already fixed it locally, but thank you!

PR #62, opened 2022-06-21
Optimize varint decode

bugadani:
It's not necessary to update a check value in every cycle - it's only relevant while processing the last encoded byte.

Obligatory godbolt link, this time with x86 asm because I was lazy: https://rust.godbolt.org/z/hMh3qW67h


jamesmunns (approved):
Works for me!

PR #64, opened 2022-06-24
Add support for 16-bit and 8-bit architectures

iFreilicht:
For example, the ATmega328P I'm using right now has an 8-bit architecture. It seems this change is all that's needed to make postcard work on that MC, but I'll add a comment once I actually tried it out.

### DIFF
--- a/src/de/deserializer.rs
+++ b/src/de/deserializer.rs
@@ -44,6 +44,18 @@ impl<'de> Deserializer<'de, Slice<'de>> {
 }
 
 impl<'de, F: Flavor<'de>> Deserializer<'de, F> {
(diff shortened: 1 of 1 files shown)


iFreilicht:
Alright, I tested it now, it works perfectly on the ATmega328P, both serialization and deserialization. I have no reason to believe it won't work for 16-bit either :)

jamesmunns:
Awesome,thank you!

As a note, although the atmegas are 8-bit CPUs, they have 16-bit pointers (and 16-bit usizes). I'm almost certain Rust wouldn't work with 8-bit pointers, but hey, I'll keep it just in case :)

Issue #46, opened 2021-10-09
pre-compute size of serialized data


x37v:
I'm using the slice flavor, I'm wondering if it is possible to compute the size of the serialized data without actually serializing?

gauteh:
Hi,

I am trying out postcard as a format for storing data on a SD-card. There will be a large number of data-packages so they will exhaust the number of files allowed on FAT32. embedded-sdmmc does not (yet) support creating directories, so I can't solve it by putting the files in dirs. So now I am considering putting many packages in each file. I will use COBS between the packages, but I need to know the serialized size of the packages. The data-package is a fixed size (no variable length fields): https://github.com/gauteh/sfy/blob/main/sfy-buoy/src/axl.rs#L10 .

- Can I compute the size of a serialized package?
- If I determine it experimentally, is it stable? Will it be the the exact same size regardless of the struct contents?

Regards, Gaute

ryankurte:
heya, i have a very similar problem and i'm not sure the macros cover the ability to pre-compute size when serializing? (or, i have missed how this can be used to achieve this).

i'm looking for something like `postcard::encoded_len<T: Serialize>(t: &T) -> usize` / equivalent to `prost`'s [`Message::encoded_len()`](https://docs.rs/prost/latest/prost/trait.Message.html#tymethod.encoded_len), that's basically a no-op encode outputting the buffer size required.

dignifiedquire:
@jamesmunns 
Any updates here? I am running into an issue with this, for preallocating buffers when serializing. The reason `MAX_SIZE` doesn't work is because I have `&[u8]` and `String` in my data. Maybe there could be a way to calculate the size based on `MAX_SIZE` + `.len()` dynamically?

jamesmunns:
Oops, forgot my "more below" part:

We COULD extend the MAX derive to take attributes on certain items, like:

```rust
#[derive(Serialize, Deserialize, MaxSize)]
struct Demo<'a> {
    arr: [Foo; 5],
    #[postcard(max_hint = 128)]
    sli: &[Bar],
    #[postcard(max_hint = 64)]
    hst: String
}
```

This covers the `!Sized` case I described above as 1b.

dignifiedquire:
@jamesmunns just created #86 which should fix things up

PR #66, opened 2022-07-01
Add function for computing the postcard serialized size of a value

lachlansneff-parallel:
This PR adds a function for computing the size of the postcard serialization of a value, as well as a ton of tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,7 @@ pub mod experimental {
     }
 
     #[cfg(feature = "experimental-derive")]
+    pub use crate::ser::serialized_size;
 
     /// Compile time Schema generation
     #[cfg(feature = "experimental-derive")]
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -279,7 +279,7 @@ where
 
 /// Compute the size of the postcard serialization of `T`.
 #[cfg(any(test, feature = "experimental-derive"))]
-pub(crate) fn serialized_size<T>(value: &T) -> Result<usize>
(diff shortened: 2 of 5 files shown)


ryankurte:
oooh this looks like it's on the way to #46?

boop @jamesmunns (plz forgive the boop-ing)

jamesmunns:
@ryankurte booping always accepted/appreciated!

jamesmunns:
Closing in favor of #66.

ChocolateLoverRaj:
> Closing in favor of #66.

@jamesmunns it seems like you made a typo in what you are referencing since you referenced this closed PR. Btw I am seeing if there is a way of "serializing" just the length of the data without serializing the data itself. Rn I could to `to_allocvec` and get the size of that, but it seems like a waste.

jamesmunns:
@ChocolateLoverRaj hmm, I'm not sure what PR I was looking for, but there is the `MaxSize` trait that you can use if your types are bounded. However, there is not a specific way to calculate the actual length without first serializing it if you have unbounded types, like `alloc/std::Vec` or `alloc/std::String` (as opposed to types like `heapless::Vec` or `heapless::String`, which have a specific upper limit).

Particularly with hand-implemented versions of the `Serialize` trait, it's difficult (or not possible) to tell what the actual serialized size is without actually exercising the serialize function.

One potential option that would still incur the same CPU effort but at least wouldn't allocate would be to implement a [serialization flavor](https://docs.rs/postcard/latest/postcard/ser_flavors/index.html) that just throws away the serialized data and just keeps a count of the bytes given to it.

PR #69, opened 2022-07-15
Make fixints usable through serde field attributes instead of wrappers

lachlansneff-parallel:
This changes the (thankfully not public, by shear accident), `fixint` module to be usable through the `#[serde(with = ...)]` field attribute instead of wrappers. 

For example:
```rust
#[derive(Serialize, Deserialize)]
pub struct DefinitelyBE {
    #[serde(with = "crate::fixint::be")]
    x: u16,
}
```


jamesmunns:
Awesome, thanks @lachlansneff-parallel!

Issue #67, opened 2022-07-13
CobsAccumulator::feed does not allow for types with references


CaptainMaso:
The accumulator struct does not allow for zero copy types that reference data stored within the accumulator.
For the following code snippet:

``` Rust
#[derive(PartialEq,Eq,serde::Serialize,serde::Deserialize)]
    struct Test<'a> {
        a : &'a str
    }

    #[test]
    pub fn main() {
        let test = Test {
            a : "This is a very long string to test the data chunking of the accumulator"
        };

        let mut data = postcard::to_allocvec_cobs(&test).unwrap();
        data.extend_from_slice(&[123,123,123,123,123,123,123,123,123,123,123,123,123,123,123,0]);
        let mut data = &data[..];

        let mut acc = crate::CobsAccumulator::<512>::new();

        let deser : Test = loop {
            let mut chunk = [0;16];
            let len = data.read(&mut chunk).unwrap();

            match acc.feed(&chunk[0..len])
            {
                postcard::accumulator::FeedResult::Consumed => continue,
                postcard::accumulator::FeedResult::OverFull(e) => println!("Full"),
                postcard::accumulator::FeedResult::DeserError(e) => println!("Error"),
                postcard::accumulator::FeedResult::Success { data, remaining } => break data,
            };
        };

        assert!(deser == test);        
    }
```

The error is:
``` Rust
error: implementation of `Deserialize` is not general enough
   --> src/lib.rs:149:23
    |
149 |             match acc.feed(&chunk[0..len])
    |                       ^^^^ implementation of `Deserialize` is not general enough
    |
    = note: `Test<'_>` must implement `Deserialize<'0>`, for any lifetime `'0`...
    = note: ...but `Test<'_>` actually implements `Deserialize<'1>`, for some specific lifetime `'1`
````

I have played around and have something that might work?
Changing the lifetime from for<'de> to an explicit 'de lifetime on the function allows the returned struct to live for as long as self is mutably borrowed, meaning we can't add more data to the accumulator until we have dropped the returned struct, and the returned struct 

i.e.
``` Rust
/// Appends data to the internal buffer and attempts to deserialize the accumulated data into
/// `T`.
pub fn feed<'a,'de, T>(&'de mut self, input: &'a [u8]) -> postcard::accumulator::FeedResult<'a, T>
where
    T: serde::Deserialize<'de>,
{
    if input.is_empty() {
        return postcard::accumulator::FeedResult::Consumed;
    }

[...]

PR #70, opened 2022-07-26
Add feed_ref to cobs_accumulator

CaptainMaso:
- Allows for deserialised types to reference data within the buffer.
   This is just a loosening of the restrictions on the deserialised
   type, so we re-use feed_ref in feed.
 - Add tests to show the use case.


jamesmunns:
Awesome! This closes https://github.com/jamesmunns/postcard/issues/67, correct? If so, could you please close that issue?

Thanks so much :)

PR #73, opened 2022-08-16
Fix Schema proc_macro to export through public experimental module.

benbrittain:
This allows Schema derive to be used by external crates


jamesmunns:
Released as `v1.0.2`.

PR #76, opened 2022-09-14
Add a Schema impl for `std::string::String` with the `use-std` feature and for `alloc::string::String` with `alloc`

zRedShift:
Hi, we found a use for `Schema` in an `std` context, and the non-`heapless` String impl was missing.

Would you be okay with adding this impl?

BTW, I find that that this experimental feature is useful for `serde` in general, not just for postcard use.


jamesmunns (approved):
Looks great, thanks!

PR #81, opened 2022-10-19
Display Custom Error Messages With `alloc` Feature

### LINKED ISSUE #80: Display Custom Error Messages
Hey there! Would you be open to displaying custom error messages when serializing and deserializing with `postman`?

I know you target resource constrained environments, so maybe it would only display the custom message if the `use-std` feature was enabled or something like that.

I'm trying to use `postman` for serializing network traffic for a video game and while writing custom serialization implementations, not being able to see my error messages makes things a little difficult at times.

zicklag:
Fixes #80.

I'm not sure if this is something you are interested in or not, but it got hard enough for me to go without it that I implemented it real quick. Let me know what you think!

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
 use core::fmt::{Display, Formatter};
 
 /// This is the error type used by Postcard
@@ -30,9 +33,9 @@ pub enum Error {
     /// The original data was not well encoded
     DeserializeBadEncoding,
     /// Serde Serialization Error
-    SerdeSerCustom,
+    SerdeSerCustom(Message),
     /// Serde Deserialization Error
-    SerdeDeCustom,
+    SerdeDeCustom(Message),
     /// Error while processing `collect_str` during serialization
     CollectStrError,
 }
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -60,32 +63,80 @@ impl Display for Error {
...
+    fn custom<T>(msg: T) -> Self
     where
         T: Display,
     {
-        Error::SerdeSerCustom
+        #[cfg(feature = "alloc")]
+        {
+            use core::fmt::Write;
+            let mut message = alloc::string::String::new();
+            write!(message, "{}", msg).expect("Write message to string");
```
iFreilicht, line 110: Hmm, I'm not particularly fond if this error message. It doesn't seem very expressive or useful. I'd say either unwrap or improve the error message to actually be useful.

src/error.rs
```diff
@@ -60,32 +63,80 @@ impl Display for Error {
...
+    fn custom<T>(msg: T) -> Self
     where
         T: Display,
     {
-        Error::SerdeDeCustom
+        #[cfg(feature = "alloc")]
+        {
+            use core::fmt::Write;
+            let mut message = alloc::string::String::new();
+            write!(message, "{}", msg).expect("Write message to string");
```
iFreilicht, line 131: Same here.

iFreilicht (commented):
I really like the concept overall, implementation is nice and the default message is very helpful.

src/error.rs
zicklag: I just decided to unwrap it. I don't think that writing to a string is supposed to fail unless it's out-of-memory or something like that, and we'd still get a backtrace on the unwrap, so I don't think a message here is bringing us much benefit.

iFreilicht (approved):
Looks good to me now. @jamesmunns, what do you think?

PR #83, opened 2022-11-25
Fix typos & encoding example in wire doc

raphlinus:
I'm pretty sure the encoding of signed 65535 is wrong, this seems like a confusion between LE and BE.

Also fixed spelling of "two's complement"


jamesmunns:
Thanks @raphlinus! I've updated the docs, and pushed a test that confirms you are correct.

PR #87, opened 2023-01-27
PhantomData's T doesn't need MaxSize impl

k3d3:
If PhantomData<T> is being used, it's very likely that the T isn't included in the data, and thus the MaxSize for it does not need to be derived.


jamesmunns:
Seems reasonable to me!

PR #90, opened 2023-02-04
fix: cobs accumulator out-of-bounds index when data is 1 byte too long

jedrzejboczar:
Currently there is an off-by-one error in bounds checking which may lead to `panic!` in `self.extend_unchecked()`. 

When calling:
```rust
let (take, release) = input.split_at(n + 1);
```
the `take` slice has length of `n + 1` ([slice::split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at)), but currently the `if` condition checks for size `n`. 

I added a test case that demonstrates the issue using `CobsAccumulator<11>` and feeding it with a 12-byte message. Without the fix this will panic:
```
thread 'accumulator::extend_unchecked_in_bounds_test' panicked at 'range end index 12 out of range for slice of length 11', library/core/src/slice/index.rs:73:5
```

### DIFF
--- a/src/accumulator.rs
+++ b/src/accumulator.rs
@@ -132,7 +132,7 @@ impl<const N: usize> CobsAccumulator<N> {
             let (take, release) = input.split_at(n + 1);
 
             // Does it fit?
-            if (self.idx + n) <= N {
+            if (self.idx + take.len()) <= N {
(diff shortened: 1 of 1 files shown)


jamesmunns:
Thank you!

Issue #95, opened 2023-03-12
Feature request: configure `fixint` at the container level


pwfff:
Maybe this is possible already and my weak Rust is the only thing stopping me, but I'm wanting to basically always use the `fixint` option, and the `with` attribute is going to be half my codebase at this point :<

If the other side of my service wasn't a completely different language I'd prefer `varint`, but for now I'd rather opt out altogether if possible.

pwfff:
I gave it a shot :)
https://github.com/jamesmunns/postcard/pull/96

stblair-crystalpeak:
This would be a great feature, I'm also in a scenario where I need to apply fixint to every single int type for a particular struct that I'm serializing. It would be so much easier if you could either apply the `with` at the struct level, or just at the container level. I hope they get to this, or accept your feature PR :)

armandas:
Need this exact feature :(

PR #96, opened 2023-03-12
Attempts at more terse `fixint` opt-ins

pwfff:
Two standalone commits here:
1. An attempt at just making the field attribute a bit more terse, but I think this is basically just the 'wrapper' y'all had before? Still have to annotate every field, plus now I have to use `into()` everywhere. Diff for my project looks like this:
```diff
diff --git a/device/src/messaging/packets.rs b/device/src/messaging/packets.rs
index 0d22758..02c80ab 100644
--- a/device/src/messaging/packets.rs
+++ b/device/src/messaging/packets.rs
@@ -1,4 +1,5 @@
 use arrayref::array_ref;
+use postcard::fixint::*;
 use serde::{Deserialize, Serialize};
 
 use super::wrapper::MAX_MESSAGE_SIZE;
@@ -34,10 +35,8 @@ pub struct RequestControllerCount {}
 
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct RequestControllerCountResponse {
-    #[serde(with = "postcard::fixint::le")]
-    packet_id: u32,
-    #[serde(with = "postcard::fixint::le")]
-    controller_count: u32,
+    packet_id: LE<u32>,
+    controller_count: LE<u32>,
 }
 
 impl Handler<RequestControllerCountResponse> for RequestControllerCount {
@@ -48,8 +47,8 @@ impl Handler<RequestControllerCountResponse> for RequestControllerCount {
 
     fn handle(&self, controller: &Controller) -> RequestControllerCountResponse {
         RequestControllerCountResponse {
-            packet_id: 0,
-            controller_count: controller.controller_count(),
+            packet_id: 0.into(),
+            controller_count: controller.controller_count().into(),
         }
     }
 }
```

2. A macro based on this post I found: https://users.rust-lang.org/t/proc-macro-attribute-to-add-a-serde-deserialize-with-to-every-field-in-a-struct/71231
This seems... fragile. I did find a serde issue/PR where they wanted to do something similar, but it's old, not very active, and had a lot of feedback I didn't quite grok. This does manage to achieve exactly what I was looking for though, so I have a copy in my own codebase that I'm proceeding with. Changes to my project look like this:
```diff
diff --git a/device/src/messaging/packets.rs b/device/src/messaging/packets.rs
index 02c80ab..b7bdeb9 100644
--- a/device/src/messaging/packets.rs
+++ b/device/src/messaging/packets.rs
@@ -1,5 +1,5 @@
 use arrayref::array_ref;
-use postcard::fixint::*;
+use corsairust_macros::all_fields_with;
 use serde::{Deserialize, Serialize};
 
 use super::wrapper::MAX_MESSAGE_SIZE;
@@ -33,10 +33,11 @@ pub trait Handler<R: Serialize> {
 
 pub struct RequestControllerCount {}
 
+#[all_fields_with("postcard::fixint::le")]
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct RequestControllerCountResponse {
-    packet_id: LE<u32>,
-    controller_count: LE<u32>,
+    packet_id: u32,
+    controller_count: u32,
 }
 
 impl Handler<RequestControllerCountResponse> for RequestControllerCount {
@@ -47,8 +48,8 @@ impl Handler<RequestControllerCountResponse> for RequestControllerCount {
 
     fn handle(&self, controller: &Controller) -> RequestControllerCountResponse {
         RequestControllerCountResponse {
-            packet_id: 0.into(),
-            controller_count: controller.controller_count().into(),
+            packet_id: 0,
+            controller_count: controller.controller_count(),
         }
     }
 }
```

There's probably a better way to make this stuff happen at the library level? Like ideally `use postcard::fixint::le::*` would just magically configure it for your whole module, but that magic is beyond me still :smiling_face_with_tear:

### DIFF
--- a/postcard-derive/src/all_fields_with.rs
+++ b/postcard-derive/src/all_fields_with.rs
@@ -1,0 +1,72 @@
+use proc_macro2::Span;
+use proc_macro2::TokenStream;
+use quote::*;
+use syn::punctuated::*;
+use syn::*;
+
+pub fn all_fields_with(
+    attr: proc_macro::TokenStream,
+    input: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+    let gen = match modify_ast_to_add_with_attr(attr.into(), ast) {
+        Ok(g) => g,
+        Err(e) => return e.to_compile_error().into(),
+    };
+    gen.into_token_stream().into()
+}
+
+fn modify_ast_to_add_with_attr(with: TokenStream, ast: DeriveInput) -> Result<impl ToTokens> {
+    let new_data: Data = match ast.data {
+        syn::Data::Struct(DataStruct {
+            struct_token,
+            fields,
+            semi_token,
+        }) => match fields {
+            fields => modify_fields_to_add_with_attr(with, struct_token, semi_token, fields)?,
+        },
+        data => data,
+    };
+    Ok(DeriveInput {
+        attrs: ast.attrs,
+        vis: ast.vis,
+        ident: ast.ident,
+        generics: ast.generics,
+        data: new_data,
+    })
+}
+
+pub fn modify_fields_to_add_with_attr(
+    with: TokenStream,
+    struct_token: syn::token::Struct,
+    semi_token: Option<syn::token::Semi>,
+    fields: Fields,
+) -> Result<Data> {
+    let mut fields = fields;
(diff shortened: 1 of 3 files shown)


pwfff:
Oh, and this is re: https://github.com/jamesmunns/postcard/issues/95

PR #97, opened 2023-03-23
improved documentation of the cobs decoding

gutzchi:
I believe that the modification of the input buffer by the cobs decoding functions is not expected as a user of the library, as it requires you to copy the input buffer every time you call it. In postcard, this does not happen with the non-cobs serde functions. 

I consider that by noting this fact in the documentation, other users might be able to realise it earlier and spend less time debugging as it happened to me.


jamesmunns:
This looks like a very reasonable addition, thanks!

PR #98, opened 2023-04-04
Introduces CRCs

huntc:
A `crc` feature has been added that brings in the `crc` crate and a new `CrcModifier` flavour. This modifier flavour is similar to the COBS implementation in how it wraps another flavour.

Convenience functions for serde are also introduced.

TODO:

* [x] Provide serialisation
* [x] Provide deserialisation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ optional = true
 
 [dependencies.paste]
 version = "1.0.12"
+optional = true
 
 [features]
 default = ["heapless-cas"]
@@ -66,7 +67,7 @@ use-std = ["serde/std", "alloc"]
 heapless-cas = ["heapless", "heapless/cas"]
 alloc = ["serde/alloc"]
 use-defmt = ["defmt"]
-use-crc = ["crc"]
+use-crc = ["crc", "paste"]
 
 # Experimental features!
 #
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,9 @@ optional = true
 version = "3.0.1"
 optional = true
 
+[dependencies.paste]
+version = "1.0.12"
+
 [features]
 default = ["heapless-cas"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,10 @@ path = "./postcard-derive"
 version = "0.1.1"
 optional = true
 
+[dependencies.crc]
+version = "3.0.1"
+optional = true
+
 [features]
 default = ["heapless-cas"]
 
@@ -59,6 +63,7 @@ use-std = ["serde/std", "alloc"]
 heapless-cas = ["heapless", "heapless/cas"]
 alloc = ["serde/alloc"]
 use-defmt = ["defmt"]
+use-crc = ["crc"]
 
 # Experimental features!
 #
--- a/src/de/flavors.rs
+++ b/src/de/flavors.rs
@@ -174,17 +174,22 @@ impl<'de> Flavor<'de> for Slice<'de> {
 ///
 /// The `crc` feature requires enabling to use this module.
 ///
-/// More on CRCs: https://en.wikipedia.org/wiki/Cyclic_redundancy_check.
-#[cfg(feature = "crc")]
+/// More on CRCs: <https://en.wikipedia.org/wiki/Cyclic_redundancy_check>.
+#[cfg(feature = "use-crc")]
 pub mod crc {
     use core::convert::TryInto;
 
     use crc::Digest;
     use crc::Width;
+    use serde::Deserialize;
 
     use super::Flavor;
+    use super::Slice;
+
+    use crate::Deserializer;
     use crate::Error;
     use crate::Result;
+    use paste::paste;
 
     /// Manages CRC modifications as a flavor.
     pub struct CrcModifier<'de, B, W>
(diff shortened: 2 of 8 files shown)


src/de/flavors.rs
huntc, line 1: Removed a test from this file as it has been superseded with a more comprehensive test for serde-ing with CRCs.

src/ser/flavors.rs
```diff
@@ -236,7 +236,7 @@ mod heapless_vec {
         }
     }
 
-    impl<'a, const B: usize> Flavor for HVec<B> {
```
huntc, line 239: Clippy discovery

src/ser/flavors.rs
```diff
@@ -380,7 +380,7 @@ where
     }
 }
 
-impl<'a, B> Flavor for Cobs<B>
```
huntc, line 383: Clippy

Cargo.toml
```diff
@@ -52,13 +52,21 @@ path = "./postcard-derive"
 version = "0.1.1"
 optional = true
 
+[dependencies.crc]
+version = "3.0.1"
+optional = true
+
+[dependencies.paste]
+version = "1.0.12"
```
huntc, line 60: For some help with our declarative macros.

src/de/mod.rs
```diff
@@ -67,6 +67,22 @@ where
...
+pub use flavors::crc::from_bytes_u32 as from_bytes_crc32;
+
+/// Conveniently deserialize a message of type `T` from a byte slice with a Crc. The unused portion (if any)
+/// of the byte slice is returned for further usage
+///
+/// See the `de_flavors::crc` module for the complete set of functions.
+/// 
+#[cfg(feature = "use-crc")]
+pub use flavors::crc::take_from_bytes_u32 as take_from_bytes_crc32;
+
```
huntc, line 85: Convenience functions. I"ve just provided the 32-bit width variants though. I figure that if the other variants are required then they can use the functions directly from the associated `flavors` package.

src/lib.rs
```diff
@@ -91,6 +91,21 @@ pub use ser::{to_stdvec, to_stdvec_cobs};
...
+
+#[cfg(all(feature = "use-crc", feature = "heapless"))]
+pub use ser::to_vec_crc32;
+
+#[cfg(all(feature = "use-crc", feature = "use-std"))]
+pub use ser::to_stdvec_crc32;
+
+#[cfg(all(feature = "use-crc", feature = "alloc"))]
+pub use ser::to_allocvec_crc32;
+
```
huntc, line 108: Same as above - conveniences for 32-bit CRCs.

tests/crc.rs
huntc, line 1: This test feels nice as to tests everything end-to-end, ensuring that the convenience functions work and serialise then deserialize correctly.

Cargo.toml
jamesmunns: Would it make sense to make this optional and gate it on the crc feature?

src/de/mod.rs
jamesmunns: That seems reasonable to me!

jamesmunns (approved):
Overall looks good to me! I don't *love* the use of macros, but I'll play around with this in a follow-on PR before releasing.

One minor comment, then should be ready to merge.

Thanks!

Cargo.toml
huntc: We could, although perhaps that adds a little complexity in case we need it elsewhere?

Alternatively, let's get rid of the macros and the need for this. :-)

huntc:
> Overall looks good to me! I don't _love_ the use of macros, but I'll play around with this in a follow-on PR before releasing.

I don't like the macros either, but I just couldn't get the `crc` crate's generics to play. Perhaps you can?

> 
> One minor comment, then should be ready to merge.

Lemme know if you insist on paste being removed as per my reply to your comment.

> 
> Thanks!

My pleasure. This library was great to work on, as well as use! Thanks for providing it!

Cargo.toml
huntc: Commit 2c9ed556ff3ce385119932e0b5c61b83519c9c62 makes the `paste` crate optional and includes it when `use-crc` is enabled.

jamesmunns:
Hey @huntc, I *think* https://github.com/jamesmunns/postcard/pull/98/commits/e437a16b9d43f79a8a10dc95e683d15f33e9bf1d was a legitimate bug, let me know if you think it looks good now.

I did try to reduce the size of the macro (so far: only on deser), but I don't think it's actually an improvement, just "not great in other ways": https://github.com/huntc/postcard/pull/1

I'm inclined to merge this as-is now, once I have your ACK on that last commit.

jamesmunns:
Part of the challenge in my generic approach is: https://github.com/mrhooray/crc-rs/issues/79.

huntc:
> Hey @huntc, I _think_ [e437a16](https://github.com/jamesmunns/postcard/commit/e437a16b9d43f79a8a10dc95e683d15f33e9bf1d) was a legitimate bug, let me know if you think it looks good now.
> 
> I did try to reduce the size of the macro (so far: only on deser), but I don't think it's actually an improvement, just "not great in other ways": [huntc#1](https://github.com/huntc/postcard/pull/1)
> 
> I'm inclined to merge this as-is now, once I have your ACK on that last commit.

Thanks James. Sorry about the bug there. Good catch!

The issues you found with generalising the width appear to be the same I had.

I’d say the main aspect of this PR is that its public API is maintainable going forward. If the underlying CRC lib was to change then I think we should remain in a good place.

PR #106, opened 2023-07-05
Fix documentation of re-exports that have dedicated doc comment

### LINKED ISSUE #65: Documentation inaccuracy in `to_std*` functions
The functions [`to_stdvec`](https://docs.rs/postcard/1.0.0/postcard/fn.to_stdvec.html) and [`to_stdvec_cobs`](https://docs.rs/postcard/1.0.0/postcard/fn.to_stdvec_cobs.html) are showing the documentation from the respective `*alloc*` functions, which means they incorrectly state that you have to enable the `alloc` feature instead of the `use-std` feature to use them.

I assume this is because the [documentation on the `pub use` definitions](https://docs.rs/postcard/1.0.0/src/postcard/ser/mod.rs.html#159) isn't picked up properly. One way to fix this would be to define this functions as `#[inline]` and call the `*alloc*` variants directly.

There might also be a bug for this on the rust-lang bugtracker, but I couldn't find one.

dtolnay:
Fixes #65.

Rustdoc's behavior for doc attributes on `pub use` items is to concatenate the attributes from the `pub use` with the attributes from the original item. It looks like that is not what was intended in the documentation in postcard.

In this PR I have converted all `pub use` that had `///` doc comments on them, into functions. This way they get to have their own self-contained documentation.

**Before:**

**After:**

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -71,17 +71,28 @@ where
 /// of the byte slice is not returned.
 ///
 /// See the `de_flavors::crc` module for the complete set of functions.
-/// 
 #[cfg(feature = "use-crc")]
-pub use flavors::crc::from_bytes_u32 as from_bytes_crc32;
+pub fn from_bytes_crc32<'a, T>(s: &'a [u8], digest: crc::Digest<'a, u32>) -> Result<T>
+where
+    T: Deserialize<'a>,
+{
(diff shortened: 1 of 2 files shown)


jamesmunns (approved):
Thanks!

jamesmunns:
Closed in favor of #107 as I didn't have permission to amend the PR. Thank you @dtolnay!

PR #111, opened 2023-08-22
Minor: Fix off-by-one in varint size calculation

Palladinium:
I was reading through the varint implementation in postcard to sanity-check one of my own implementations, and I spotted a mistake in the code that calculates the maximum number of varint bytes an integer datatype can take.

The old version adds 7 instead of 6, so instead of the behaviour documented in the comment it would overestimate the number of varint bytes required by one for datatypes whole size is a multiple of 7 bits. Since there are thankfully none among the common integer datatypes, the behaviour is correct both before and after the change.


jamesmunns:
Thanks, great catch!

PR #115, opened 2023-11-14
Update heapless

luqasz:


dzamlo:
This would be a breaking changes.

jamesmunns:
Yeah, I agree with @dzamlo, heapless types appear in the public interface, so upgrading it would be a breaking change.

I am not certain, but I've considered doing the following:

* adding `heapless-v0_7` and `heapless-v0_8` features, using those versions specifically, maybe in some module-d scope, for things like `to_vec`, `to_vec_cobs`
* keeping `heapless` as an alias for `heapless-v0_7`, but marking it as deprecated.

Open to feedback if people think there is a preferrable solution.

jamesmunns:
Merging and will sort out the versioning stuff before the next release.

Issue #118, opened 2023-11-16
Support latest version of `embedded-io` 


Anders429:
I see that #91 added support for `embedded-io`'s `Read` and `Write` traits. However, the support is only for version `0.4`. Since `embedded-io` is now on version `0.6`, is it possible for `postcard` to increase the version of its dependency accordingly?

Issue #121, opened 2024-01-04
Decide how to deal with heapless-v0.8


jamesmunns:
heapless types appear in the public interface, so upgrading it would be a breaking change.

I am not certain, but I've considered doing the following:

* adding `heapless-v0_7` and `heapless-v0_8` features, using those versions specifically, maybe in some module-d scope, for things like `to_vec`, `to_vec_cobs`
* keeping `heapless` as an alias for `heapless-v0_7`, but marking it as deprecated.

Open to feedback if people think there is a preferrable solution.

_Originally posted by @jamesmunns in https://github.com/jamesmunns/postcard/issues/115#issuecomment-1877214683_

bsodmike:
Hi James,

Latest postcard has a dependency on "heapless 0.7.17" which depends on `atomic-polyfill`.  I'm trying to upgrade my dep tree to move to `portable_atomic` supports `AtomicU64`. Do you think you can help with this please?

Cheers

enbyted:
Hi,

In my view the best approach would be to get rid of other crates in the public API, as you proposed in #128 . However, I assume that releasing 2.0 might take a while, so as a temporary solution doing what you proposed here (having the 2 features flags) sounds reasonable.

In one of my projects I've used heapless 0.8 for quite a while before deciding on using postcard, which means that I either need to do a breaking downgrade now or implement my own heapless (0.8) compatibility for postcard. I guess I could also depend on postcard master, but I'm not really keen on doing so.

jamesmunns:
Someone has made an adapter for implementing `to_vec` using heapless v0.8: https://github.com/elrom/postcard-heapless-compat.

At some point in the future, I plan to add impls for both heapless v0.7 and v0.8 behind feature flags, and deprecating the top level methods to avoid confusion. The idea would be to then keep that shape (and required feature flags) for postcard v2.0.

Gerharddc:
@jamesmunns thanks! I'm not sure how to use that adapter, but simply creating and using a fork of postcard that uses heapless-v0.8 seems straightforward enough. It appears that nothing inside postcard actually breaks with the update. Perhaps this could be considered as a change for postcard 1.2.0 (as in https://github.com/jamesmunns/postcard/pull/211)?

jamesmunns:
As mentioned at the top of this issue:

> heapless types appear in the public interface, so upgrading it would be a breaking change.

PR #123, opened 2024-01-11
#118 Add support for latest embedded-io

elast0ny:
This adds a new `embedded-io-06`  feature which allows users to use the latest `embedded_io` v `0.6.*`.

Backwards compatibility is maintained with the previous `embedded-io` which is now setup to activate the `embedded-io-04` feature internally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "postcard"
-version = "1.0.8"
+version = "1.0.9"
 authors = ["James Munns <james@onevariable.com>"]
 edition = "2018"
 readme = "README.md"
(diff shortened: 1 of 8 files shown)


jamesmunns:
Thank you @elast0ny! I'm not 100% I want to use the pattern you did, of mutually exclusive features. I may want to instead offer both, in separate modules, with the existing `embedded-io` meaning `0.4`, and mark it as deprecated.

See also #115 and #121. I'll want to be consistent with whatever I decide for the version bump.

elast0ny:
Thanks for the quick reply ! I needed support for v`0.6` so I'll probably go on and use my fork given that this depends on some design decisions and #115 / #121  being resolved.

Let me know when you have a concrete path forward for this and I'll see if I can update my branch/PR !

jamesmunns:
I'm going to merge this and then move it around a bit, thank you @elast0ny!

elast0ny:
No problem ! Just as an FYI, I've opened https://github.com/rust-embedded/embedded-hal/issues/566 to see if we cant get a "stable" `embedded-io` release. It would be nice if the ecosystem would converge to a `1.0.0` that will hopefully stick around for a long time.

Issue #128, opened 2024-01-31
RFC: Thoughts on Postcard 2.0?


jamesmunns:
Due to some of the dependency stuff, I am considering doing a 2.0 **breaking change** to the postcard crate. This would allow me to make public API changes to make dependencies, particularly for those that are required for `flavors`, like `heapless`, or `embedded-io`, to be versioned, so that updating them does not cause future breaking changes.

This would **not** be a breaking change to the wire format: Postcard 1.x and Postcard 2.x users would be able to mutually send messages with no compatibility issues, but there WOULD be breaking API changes, so some code changes could be required for updating.

As I see it, the options today are:

1. Don't do a breaking change, but add support for new version of libraries like heapless and embedded-io
    * This will likely be awkward, and have complicated feature flags
    * I'd probably deprecate certain APIs (like `to_vec()` that is for heapless v0_7)
    * I don't really want to spend time on this.
2. Do a breaking change to the LIBRARY API, but not the BINARY WIRE FORMAT.
    * This means we can't fix issues like #101, but we can fix the other stuff
    * I'll probably isolate the "flavor" impls better, either as feature-flagged modules, or maybe even as external crates.
    * This way people can update when they want, and if they really want to backport flavors (like for heapless-v0_8) to postcard 1.0, they can.
    * This is my preferred option
3. Break both the library API AND the binary format
    * this would let me fix the char issue, but we don't actually have any other open issues AFAIK
    * this would be unfortunate, and FORCE users to update, and might cause compat issues.

If you use postcard, I'd like to hear your opinion! Unless someone is willing to do the work for option 1, or willing to sponsor the work for 1, or makes a suggestion for another option, I'll probably go with 2.

Tagging some related issues:

* #124
* #123 
* #126
* #121
* #103

CBJamo:
The cost of option 2 seems very low to users. Each project I update would only be looking at 2-5 lines to change, and the changes would be very similar from project to project. If it makes maintenance easier and opens the door to more features it seems like an obviously good choice to me.

The cost of option 3 is higher, but the cost of making that change is only ever going to go up. This seems to me like a "best time to plant a tree" type situation. Personally, when I make a change around postcard it's because I'm changing the data. Since that means changing both sides of the communication, there isn't actually any cost to a wire protocol change. On the other hand I don't think I've ever sent a char with postcard, so the char issue doesn't actually seem like that much of an issue.

jamesmunns:
[...]

PR #129, opened 2024-01-31
Remove unnecessary lifetimes of CRC digest

cdunster:
The CRC digest lifetime is set to the same as the provided buffer lifetime which means that the CRC has to live as long as the buffer which I don't think is necessary as the `finalize()` method is called on the digest. This strictness of the lifetime meant that I couldn't create a `Crc` in the local scope i.e. a method and then just return the bytes (see the new unit test for an example of what I mean).

Clippy suggests this change and to remove a few other lifetimes too so I think this should be a low risk change :slightly_smiling_face:

Thanks for the great crate!

### DIFF
--- a/src/ser/flavors.rs
+++ b/src/ser/flavors.rs
@@ -638,10 +638,10 @@ pub mod crc {
                     ///
                     /// When successful, this function returns the slice containing the
                     /// serialized and encoded message.
-                    pub fn [<to_slice_ $int>]<'a, 'b, T>(
-                        value: &'b T,
+                    pub fn [<to_slice_ $int>]<'a, T>(
+                        value: &T,
                         buf: &'a mut [u8],
-                        digest: Digest<'a, $int>,
+                        digest: Digest<'_, $int>,
                     ) -> Result<&'a mut [u8]>
                     where
                         T: Serialize + ?Sized,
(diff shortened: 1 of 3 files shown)


jamesmunns:
CC @huntc - do you know if these lifetimes were chosen intentionally?

jamesmunns:
Also, I need to figure out if removing a lifetime like this is a breaking API change. If yes, it might need to wait for #128, but I don't think it is?

huntc:
I've just pulled down the latest commit on main, updated my toolchain to the latest (`rustc 1.75.0 (82e1608df 2023-12-21)
`), and performed a `cargo clippy`.

All I see that clippy recommends is the following:

```
warning: use of `default` to create a unit struct
   --> src/ser/serializer.rs:377:50
    |
377 |                     .map_err(|_| core::fmt::Error::default())
    |                                                  ^^^^^^^^^^^ help: remove this call to `default`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#default_constructed_unit_structs
    = note: `#[warn(clippy::default_constructed_unit_structs)]` on by default

warning: the following explicit lifetimes could be elided: 'a
   --> src/ser/mod.rs:269:18
    |
269 | pub fn to_extend<'a, T, W>(value: &'a T, writer: W) -> Result<W>
    |                  ^^                ^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
    = note: `#[warn(clippy::needless_lifetimes)]` on by default
help: elide the lifetimes
    |
269 - pub fn to_extend<'a, T, W>(value: &'a T, writer: W) -> Result<W>
269 + pub fn to_extend<T, W>(value: &T, writer: W) -> Result<W>
```

The explicit lifetime for `to_extend` is not required. However, I don't see that in this PR.

Regarding the lifetimes on `Digest`, the `_` lifetime should be fine. I don't think this would be a breaking a change as the lifetime there relates to the Crc. I'm not sure why I originally associated the lifetime with the buffer.

huntc:
Should we progress this pull request?

jamesmunns:
Thank you all!

PR #130, opened 2024-03-01
Depend directly on serde_derive

hanna-kruppe:
Since [serde v1.0.186](https://github.com/serde-rs/serde/releases/tag/v1.0.186), this is possible without risking errors from Cargo picking incompatible versions of `serde` and `serde_derive`. The benefit of *not* enabling serde's "derive" feature is that (non-incremental) builds can become a bit faster, as serde itself (and some things depending on it) can be compiled in parallel to proc-macro2, syn, and `serde_derive`. See https://github.com/serde-rs/serde/issues/2584 for more discussion and background. In a project of mine where all transitive dependencies except postcard have already made the switch, this patch makes a clean build 3-4 seconds faster (out of a minute in total) on my machine.

The downside of this change, besides the additional `use` lines, is that it's easy to accidentally write code that breaks when used in a project where something else enables serde's "derive" feature. Importing the traits from `serde::{Deserialize, Serialize}` works until it doesn't, one has to remember to always import the traits from `serde::ser` and `serde::de` (see https://github.com/serde-rs/serde/issues/1441). I think I did it correctly in this PR, but there's no good way to test it other than occasionally editing Cargo.toml to re-enable serde's "derive" feature. I'm honestly not sure if I'd want to deal with that as a library maintainer, no hard feelings if you don't want to either.

The second commit goes one step further and makes postcard's `serde_derive` dependency completely optional except as a dev-dependency, since it only seems to be needed for tests and for postcard-derive-related code in schema.rs. This is less impactful since virtually every crate depending on postcard also needs `serde_derive` anyway, but it shouldn't hurt either and doesn't make things any more complicated than the first commit.

This is a breaking change: projects using postcard are likely to depend on serde re-exporting the derive macros, and for some of them, postcard will be the only thing enabling serde's "derive" feature. But since a 2.0 release is likely coming anyway, I wanted to pitch this change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,9 +35,11 @@ features = ["serde"]
 optional = true
 
 [dependencies.serde]
-version = "1.0.100"
+version = "1.0.186"
 default-features = false
-features = ["derive"]
+
+[dependencies.serde_derive]
+version = "1"
 
 [dependencies.cobs]
 version = "0.2.3"
@@ -69,6 +71,9 @@ optional = true
 [dependencies.paste]
 version = "1.0.12"
 optional = true
+
+[dev-dependencies.serde_derive]
+version = "1"
 
 [features]
 default = ["heapless-cas"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ default-features = false
 
 [dependencies.serde_derive]
 version = "1"
+optional = true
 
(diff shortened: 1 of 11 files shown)


dtolnay (commented):
As of https://github.com/jamesmunns/postcard/pull/219 postcard no longer depends on serde derive.

PR #134, opened 2024-04-01
Fix #89: `#[derive(Schema)]` for `StructVariant`s

### LINKED ISSUE #89: `#[derive(Schema)]` not working with `StructVariant`s
The following fails:

```rust
#[derive(Serialize, Deserialize, MaxSize, Schema)]
enum Example {
    VarA,
    VarB(u32),
    VarC {
        d: u8,
        e: u16,
    }
}
```

with

```
error[E0308]: mismatched types
  --> src/main.rs:45:9
   |
45 |         d: u8,
   |         ^
   |         |
   |         expected `&NamedValue`, found struct `NamedValue`
   |         help: consider borrowing here: `&d`
```

The derive macro is likely not handling struct variants correctly at the moment.

barafael:
Here comes a 4-commit sandwich.

The actual issue I believe is fixed with this. However, there are 2 open questions regarding the schema tests (see commit message bodies):

* Where should the schema tests be located? As of now, they force the schema module to be `pub`lic because they are integration tests. A doc comment above `mod schema;` explicitly says not to make experimental features `pub` :laughing:
* The test for struct serialization does not pass. It seems it wants the array member to be schematized as a `Seq`, but the array member is schematized as a 10-element `Tuple`.

Let me know what you think :) I love using this crate.

### DIFF
--- a/postcard-derive/src/schema.rs
+++ b/postcard-derive/src/schema.rs
@@ -89,7 +89,7 @@ fn generate_variants(fields: &Fields) -> TokenStream {
             let fields = fields.named.iter().map(|f| {
                 let ty = &f.ty;
                 let name = f.ident.as_ref().unwrap().to_string();
-                quote_spanned!(f.span() => ::postcard::experimental::schema::NamedValue { name: #name, ty: <#ty as ::postcard::experimental::schema::Schema>::SCHEMA })
+                quote_spanned!(f.span() => &::postcard::experimental::schema::NamedValue { name: #name, ty: <#ty as ::postcard::experimental::schema::Schema>::SCHEMA })
             });
             quote! { &::postcard::experimental::schema::SdmTy::StructVariant(&[
                 #( #fields ),*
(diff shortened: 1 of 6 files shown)


jamesmunns:
Hey @barafael, overall looks good!

1. can we use the scoped/conditional exports, e.g. `postcard::experimental::schema` instead of making `postcard::schema` public? If not, could you please slap it with a `#[doc(hidden)]` and a comment that says "this isn't API stable, just an impl detail"?
2. Re: https://github.com/jamesmunns/postcard/pull/134/commits/5d0e703feef85eda59e56a7037af503a2afdb87f, I dunno, I'll have to look, that looks very seq-y to me.

Thanks!

barafael:
1.: maybe a simple conditional `pub` is good enough?

jamesmunns:
@barafael I fixed some merge conflicts, and looked at the test.

I believe the test as-is now is correct: fixed size arrays, such as `[T; N]`, should be serialized as tuples, as the length is known, so the length is not sent on the wire. Slices (and Vecs) such as `&[T]` should be serialized as `Seq`s, as their length WILL be sent on the wire.

Thank you!

PR #137, opened 2024-04-04
Define MaxSize for Arc and Rc

carlsverre:
Serde supports auto-deriving Serialize and Deserialize for Arc and Rc types with the caveat being that the underlying type is serialized/deserialized. I am using this feature along with Postcard, and need this patch (or something equivalent to it) in order to calculate the max serialized size of my types.

More details on Serde's Arc and Rc support here: https://serde.rs/feature-flags.html#rc

### DIFF
--- a/src/max_size.rs
+++ b/src/max_size.rs
@@ -1,3 +1,9 @@
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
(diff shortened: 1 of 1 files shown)


jamesmunns:
Thank you!

Issue #144, opened 2024-05-13
This repository doesn't seem to have continuous integration


ia0:
This crate already has quite a few dependents but doesn't have any form of continuous integration as far as I can tell. This seems a bit risky and makes this crate looks less reliable. In particular, in #142 I realized that the `tests/schema.rs` could never have been run because of the wrong `cfg(feature = "")` gating the whole file. I understand this feature is experimental, but I would expect a repository to contain a few dead-code as possible, including tests.

Are there any plans of adding continuous integration?

jamesmunns:
@ia0 thank you for all the support today!

PR #145, opened 2024-05-13
Fix compilation with alloc and experimental-derive

ia0:
```sh
RUSTFLAGS='--check-cfg=cfg(doc_cfg)' cargo check --features=alloc,experimental-derive
```

Note that with a continuous integration, this would probably not happen, see #144.


jamesmunns (approved):
Thank you! Still open to a PR for CI.

PR #146, opened 2024-05-13
Use well-known docsrs cfg for docs.rs

ia0:
See https://github.com/rust-lang/cargo/pull/13383 for the PR adding support.

Without this change, there are many warnings about https://blog.rust-lang.org/2024/05/06/check-cfg.html when compiling.


jamesmunns (approved):
Thank you!

PR #150, opened 2024-05-13
Fix clippy lints

ia0:
I tried to choose the most obvious fix for each warning, but sometimes it might be arguable. I'll try to point out those cases. Feel free to suggest alternatives if needed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,11 +17,8 @@ mod varint;
 pub(crate) mod max_size;
 
 /// The schema types and macros.
-#[cfg(not(feature = "experimental-derive"))]
-mod schema;
-
 #[cfg(feature = "experimental-derive")]
-pub mod schema;
+pub(crate) mod schema;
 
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -17,11 +17,8 @@ mod varint;
 pub(crate) mod max_size;
 
 /// The schema types and macros.
-#[cfg(not(feature = "experimental-derive"))]
-mod schema;
-
 #[cfg(feature = "experimental-derive")]
-pub mod schema;
+pub(crate) mod schema;
```
ia0, line 21: This is partially reverting #134 because I don't think the intent was to expose `schema` publicly until it goes out of experimental. The warning here was about documentation, which should be duplicated if we want to control the visibility from the `cfg`.

jamesmunns: I will likely intend to take this out of experimental with #128 coming up soon, so it works for me.

PR #151, opened 2024-05-13
Relayout the repo

jamesmunns:
This is the first step towards breaking out dependencies into smaller crates

PR #108, opened 2023-07-19
Render cfg information on docs.rs

dtolnay:
Screenshot from 2023-07-18 16-59-08

Screenshot from 2023-07-18 17-00-34

PR #152, opened 2024-05-13
Fix documentation lints

ia0:


ia0:
This actually depends on #150 to pass CI.

jamesmunns:
CC @ia0, I'm not actually sure what the failing line is supposed to do, it was introduced in https://github.com/jamesmunns/postcard/pull/108 by @dtolnay, it's possible this is only intended to be activated in docs-rs with a nightly build, if so should we avoid activating this in CI?

ia0:
The last commit should fix it. We should indeed only test the documentation in nightly. Also regarding the `doc_cfg` feature, we could probably simplify the code by using `auto_doc_cfg` instead. This could be done in another PR (after #151).

PR #154, opened 2024-05-14
Add #[postcard(bound = "...")] attribute for derive(Schema)

### LINKED ISSUE #153: derive(Schema) introduces incorrect bounds instead of doing perfect derive
This is a [known issue](https://github.com/rust-lang/rust/issues/26925) in Rust that `derive(Clone)`, `derive(PartialEq)`, etc., don't infer the proper bounds. They will require each type parameter to implement the trait instead of looking at what is actually needed by the type definition (called [perfect derive](https://smallcultfollowing.com/babysteps/blog/2022/04/12/implied-bounds-and-perfect-derive/)).

For some reason, `serde` derive macros do perfect derive. It would be nice if `Schema` derive macro would also do so. Here is a small reproduction example:

```toml
# Cargo.toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

[dependencies]
postcard = { version = "1.0.8", features = ["experimental-derive"] }
serde = { version = "1.0.201", features = ["derive"] }

[patch.crates-io.postcard]
git = "https://github.com/jamesmunns/postcard.git"
rev = "4e1104d9e2e1ef6f9b2473cf5da7d9d58fd8f6fd"
```

```rust
// src/lib.rs
use postcard::experimental::schema::Schema;

pub trait View<'a> {
    type Type<T: Wire<'a>>: Wire<'a>;
}

#[derive(Serialize, Deserialize, Schema)]
pub enum Api<'a, T: View<'a>> {
    Foo(T::Type<Foo<'a>>),
}

#[derive(Serialize, Deserialize, Schema)]
pub struct Foo<'a> {
    buf: &'a [u8],
    ret: u32,
}

pub enum Id {}
impl<'a> View<'a> for Id {
    type Type<T: Wire<'a>> = T;
}

pub trait Wire<'a>: Serialize + Deserialize<'a> + Schema {}
impl<'a, T: Serialize + Deserialize<'a> + Schema> Wire<'a> for T {}

#[test]
fn api_schema() {
    let schema = Api::<Id>::SCHEMA;
}
```

Running `cargo test` gives the following error:

```
error[E0599]: the variant or associated item `SCHEMA` exists for enum `Api<'_, Id>`, but its trait bounds were not satisfied
   --> src/lib.rs:29:29
    |
9   | pub enum Api<'a, T: View<'a>> {
    | ----------------------------- variant or associated item `SCHEMA` not found for this enum because it doesn't satisfy `Api<'_, Id>: Schema`
...
19  | pub enum Id {}
    | ----------- doesn't satisfy `Id: Schema`
...
29  |     let schema = Api::<Id>::SCHEMA;
    |                             ^^^^^^ variant or associated item cannot be called on `Api<'_, Id>` due to unsatisfied trait bounds
    |
    = note: trait bound `Id: Schema` was not satisfied
    = note: the following trait bounds were not satisfied:
            `Api<'_, Id>: Schema`
            which is required by `&Api<'_, Id>: Schema`
```

In particular, it thinks that `Id` must implement `Schema`, while only `Id::Type<Foo<'a>>` (i.e. `Foo<'a>`) needs to.

When looking at the `cargo expand` output, we can see the difference between `serde::Serialize` and` postcard::Schema`:

```
    #[automatically_derived]
    impl<'a, T: View<'a>> _serde::Serialize for Api<'a, T>
    where
        T::Type<Foo<'a>>: _serde::Serialize, // This is the correct bound.
    { ... }
// versus
impl<
    'a,
    T: View<'a> + ::postcard::experimental::schema::Schema, // This is an incorrect bound.
> ::postcard::experimental::schema::Schema for Api<'a, T> { ... }
```

This issue is blocking #143.

ia0:
This doesn't fix #153 but provides a work-around (which is also present in `serde`).


max-heller:
Hi @ia0, I merged this with main since things have changed quite a bit since you opened the PR. Could you could add a section to the docs [here](https://github.com/jamesmunns/postcard/blob/493ae45406d49d310e90cec272cf29904bb85c3e/source/postcard-schema/src/lib.rs#L23) describing the attribute?

@jamesmunns have you thought any more about the future of `MaxSize`? Currently, this only affects `derive(Schema)`, but if we want to keep `MaxSize` as a derive it should probably take this attribute into account or have a more fine-grained version (`#[postcard(bound(schema = "...", max_size = "..."))]`?)

ia0:
> Could you could add a section to the docs [here](https://github.com/jamesmunns/postcard/blob/493ae45406d49d310e90cec272cf29904bb85c3e/source/postcard-schema/src/lib.rs#L23) describing the attribute?

Yes. However, note that I probably won't be able to spend too much time on this PR since I ended up writing and using my own serialization library [`wasefire-wire`](https://docs.rs/wasefire-wire).

max-heller:
> However, note that I probably won't be able to spend too much time on this PR since I ended up writing and using my own serialization library [`wasefire-wire`](https://docs.rs/wasefire-wire).

I think this is pretty much good to go, thanks for the contribution and the pointer to `wasefire` which looks quite interesting!

jamesmunns:
@max-heller I haven't had a chance yet! I may have some time on postcard over the next two weeks. I'm inclined to favor the const fn version, but I know there are open design points on that as well right now. I wouldn't object to extending this change to MaxSize, but also fine if it stays as-is for the moment.

@ia0 I'm not sure if you've done a side-by-side with wasefire and postcard-rpc, I'd definitely be interested to help set one up to understand the differences in design you ended up with! I remember the use cases being pretty overlapping? Happy to coordinate in an issue or email if you think it would be useful.

ia0:
> I'm not sure if you've done a side-by-side with wasefire and postcard-rpc, I'd definitely be interested to help set one up to understand the differences in design you ended up with! I remember the use cases being pretty overlapping? Happy to coordinate in an issue or email if you think it would be useful.

No I didn't do anything precise because wasefire is much more opinionated (wasefire-wire is not really meant to be used outside wasefire even if the wire format is postcard compatible). The only question is whether postcard-rpc covers what wasefire-protocol needs. I don't know how much the schema changed since last time I checked, but that was the main issue. There was also things I didn't like about serde, but I don't remember how much of an issue it was. It's true that it would be good to check this at some point. I've created https://github.com/google/wasefire/issues/708 but probably won't be able to look at it before late 2025.

PR #155, opened 2024-05-15
Remove unnecessary lifetime from CRC digest in deserializer

MathiasKoch:
This PR is the deserialize complement to https://github.com/jamesmunns/postcard/pull/129

This is required to allow wrappers such as

```rust
fn deserialize(buf: &[u8]) -> Result<Data<'_>, Error> {
    let crc = Crc::<u32>::new(&CRC_32_ISCSI);
    let digest = crc.digest();
    postcard::from_bytes_crc32(buf, digest).map_err(|_| Error::Deserialize)
}
```

without getting
```
   |
54 |         let digest = crc.digest();
   |                      --- `crc` is borrowed here
55 |         postcard::from_bytes_crc32(buf, digest).map_err(|_| Error::Deserialize)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

### DIFF
--- a/source/postcard/src/de/mod.rs
+++ b/source/postcard/src/de/mod.rs
@@ -100,7 +100,7 @@ where
 #[cfg(feature = "use-crc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "use-crc")))]
 #[inline]
-pub fn from_bytes_crc32<'a, T>(s: &'a [u8], digest: crc::Digest<'a, u32>) -> Result<T>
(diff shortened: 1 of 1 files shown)


MathiasKoch:
Nevermind this.. Moved a bit too fast with the PR.. Don't think this will work in practice?

PR #157, opened 2024-05-28
Do not discard error messages when using std

### LINKED ISSUE #38: Errors are not properly propagated
I forgot to implement `visit_seq` for deserializing a struct. That caused postcard to fail deserialize it but json-serde succeeded as it users `visit_map`. The problem is that for any error which happens during serialization postcard returns the error `SerdeDeCustom`. 

```
called `Result::unwrap()` on an `Err` value: SerdeDeCustom
thread 'tests::integration::test_serialisation_postcard' panicked at 'called `Result::unwrap()` on an `Err` value: SerdeDeCustom', src/tests.rs:74:88
```

If I do not implement `visit_seq` this code is called by serde:

```
    fn visit_seq<A>(self, seq: A) -> Result<Self::Value, A::Error>
    where
        A: SeqAccess<'de>,
    {
        let _ = seq;
        Err(Error::invalid_type(Unexpected::Seq, &self))
    }
```

The error ` Err(Error::invalid_type(Unexpected::Seq, &self))` just vanishes and becomes a  `SerdeDeCustom`. This is very hard to debug.

### LINKED ISSUE #80: Display Custom Error Messages

HalfVoxel:
Fixes #38
Fixes #80

### DIFF
--- a/source/postcard/src/error.rs
+++ b/source/postcard/src/error.rs
@@ -31,8 +31,16 @@ pub enum Error {
     DeserializeBadEncoding,
     /// Bad CRC while deserializing
     DeserializeBadCrc,
+    #[cfg(feature = "use-std")]
+    /// Serde Serialization Error
+    SerdeSerCustom(String),
+    #[cfg(not(feature = "use-std"))]
     /// Serde Serialization Error
     SerdeSerCustom,
+    #[cfg(feature = "use-std")]
+    /// Serde Deserialization Error
+    SerdeDeCustom(String),
+    #[cfg(not(feature = "use-std"))]
     /// Serde Deserialization Error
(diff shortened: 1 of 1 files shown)


jamesmunns:
I'm pretty sure this is a breaking change as it changes the error variant. I'll likely include this as part of postcard 2.0.

Issue #167, opened 2024-08-21
Added Dependency on Arc


sjudson:
The 1.0.9 release of `postcard` appears to have added (through https://github.com/jamesmunns/postcard/commit/ae7f40b21533ef314d789ae18f5c45508d2a3c32) a dependency for the `alloc` feature on `alloc::sync::Arc`. 

Per the Rust docs (https://doc.rust-lang.org/alloc/sync/index.html), this is unsupported on certain targets:

> Note: This module is only available on platforms that support atomic loads and stores of pointers. This may be detected at compile time using #[cfg(target_has_atomic = "ptr")].

Would it be possible to gate the use of `Arc` against the `target_has_atomic` configuration check recommended by the docs?

I'd be happy to open a PR to do so if that would be helpful.

jamesmunns:
@sjudson I would appreciate a PR, yes! Feel free to @ me when it is open.

sjudson:
> @sjudson I would appreciate a PR, yes! Feel free to @ me when it is open.

Great, will do.

jamesmunns:
Released as v1.0.10, thank you @sjudson!

https://github.com/jamesmunns/postcard/releases/tag/v1.0.10

PR #168, opened 2024-08-21
Check that target supports atomic pointer operations before defining max size for Arc

sjudson:
Resolving #167 cc: @jamesmunns 

When building with, _e.g._, `--target riscv32i-unknown-none-elf` and `-F alloc` on the current main branch, the build will fail with 

```
error[E0432]: unresolved import `alloc::sync`
 --> source/postcard/src/max_size.rs:5:33
  |
5 | use alloc::{boxed::Box, rc::Rc, sync::Arc};
  |                                 ^^^^ could not find `sync` in `alloc`
```

due to the lack of support on that target for atomic pointer operations (see: https://doc.rust-lang.org/alloc/sync/index.html).

Following the recommendation from the Rust docs, this gates definition of `POSTCARD_MAX_SIZE` for `alloc::sync::Arc` behind a configuration check to confirm it is available on the target first.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -9,6 +9,7 @@ ensure_target() {
 }
 
 ensure_target thumbv7em-none-eabi
+ensure_target riscv32i-unknown-none-elf
 
 has_toolchain() {
   rustup toolchain list | grep -q "$1"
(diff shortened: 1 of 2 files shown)


jamesmunns:
Gotcha! And you can confirm this resolves the issue you saw in #167? If so, I'll merge and release after CI is green.

jamesmunns:
Also @sjudson if you'd like to add a test for this arch (rv32 + alloc) to the CI so it doesn't regress in the future: https://github.com/jamesmunns/postcard/blob/main/ci.sh feel free to add it to this PR!

sjudson:
Thanks!

PR #170, opened 2024-09-06
Implement a version of Schema types that are owned

jamesmunns:
These types are use the heap to handle the recursive nature of schemas, and can be punned or converted with existing typed or serialized schema types.

For now, the serialized form is fairly inefficient: types and names are repeated instead of being interned or deduplicated in any way.


jamesmunns:
Merging this, I am going to split out the schema stuff to it's own crate

PR #171, opened 2024-09-06
chore: pin exact version of `heapless`

kaffarell:
For the moment the `heapless` crate is locked on 0.7.0, because 0.8.0 still requires some work (the revert happened with #164). The problem is that the "0.7.0" version requirement also matches "0.8.0", so for the moment we need "~0.7.0" to not allow a minor version update (which would brake the package).

### DIFF
--- a/source/postcard/Cargo.toml
+++ b/source/postcard/Cargo.toml
@@ -25,7 +25,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 
 [dependencies.heapless]
-version = "0.7.0"
+version = "~0.7.0"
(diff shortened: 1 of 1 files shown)


jamesmunns:
Can you clarify what you think this changes? As far as I am aware, with how cargo resolves semver versions, `0.7.0` will not ever resolve to `0.8.0`, even without the `~`.

When the major version is `0`, a minor version change is [treated as a major change](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility)

kaffarell:
You're right, this was an error on my local registry, my bad!

PR #173, opened 2024-09-15
Corrected invalid link in `Serializer` docs

TylerBloom:
There is a broken link in the docs for the `Serializer`. The intended link seems to be to `crate::ser_flavors::Flavor`.


jamesmunns (approved):
Thanks!

PR #174, opened 2024-09-17
Add support for skipped fields

dns2utf8:
Hi all

I made this PR to support maybe skipped fields, either always or dynamically with `#[serde(skip_serializing_if = "Option::is_none")]`.

I have considered hiding the function behind a feature flag and adding a runtime Err without the flag so consumers of the crate notice that they were trying to serialize an object that would not be readable again.

Looking forward hearing from you,
Stefan

PS: I assume that `None::<()>` takes the same amount of space as `None::<u128>` or something else would and that trick then allows us to stay with the wire format, 99% sure of that, happy to learn :blush:

### DIFF
--- a/source/postcard/src/ser/serializer.rs
+++ b/source/postcard/src/ser/serializer.rs
@@ -521,6 +521,11 @@ where
     }
 
     #[inline]
+    fn skip_field(&mut self, _key: &'static str) -> Result<()> {
+        None::<()>.serialize(&mut **self)
+    }
+
+    #[inline]
     fn end(self) -> Result<()> {
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


ManevilleF:
Does this also fix the `#[serde(skip)]` attribute ? or is it only targetted to `skip_serializing_if` ?

dns2utf8-novaziun:
It works with `skip`, but it serialized it to a `None`, so I am not sure what happens if your type is not `Option<T>`

max-heller:
This works for (conditionally) skipped fields of type `Option<T>`, but I don't think it will work for fields with different types, for skipped variants, or for skipped fields of tuple structs/variants (since those don't have `skip_field()` methods). Unfortunately, I don't see a way to handle skipped fields in their full generality given `serde`s current APIs without serializing field names or indices, which would break the wire format and add overhead to structs with no skipped fields.

jamesmunns:
I'm not sure this will work in the general case, so I'm going to close this. Thank you @dns2utf8 for submitting this.

PR #175, opened 2024-09-29
Implement Schema for nalgebra matrix type

avsaase:
I want to send some nalgebra types over the wire using postcard-rpc so I needed to implement the `Schema` trait. Initially I used a newtype to get around the orphan rule but in this case that's quite annoying because it requires copying a _huge_ API into my own project.

I figured `nalgebra` is a somewhat foundational crate in the Rust ecosystem to I think it is acceptable to add it an an optional dependency here.

### DIFF
--- a/source/postcard/Cargo.toml
+++ b/source/postcard/Cargo.toml
@@ -69,6 +69,7 @@ optional = true
 [dependencies.nalgebra]
 version = "0.33.0"
 optional = true
+default-features = false
 
 [features]
 default = ["heapless-cas"]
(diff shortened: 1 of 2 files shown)


source/postcard/Cargo.toml
```diff
@@ -66,6 +66,11 @@ optional = true
 version = "1.0.12"
 optional = true
 
+[dependencies.nalgebra]
+version = "0.33.0"
```
jamesmunns, line 70: Since nalgebra isn't 1.0, could you please make this an explicit feature with a name like `nalgebra-0_33`? This will prevent this from being an issue in the future if we need to support multiple versions.

avsaase: Done

avsaase:
Thanks for including this. And nice work on the update!

PR #179, opened 2024-10-31
MaxSize as a const fn

jamesmunns:


source/postcard-schema/src/max_size.rs
```diff
@@ -0,0 +1,122 @@
...
+            while i < nts.len() {
+                let Some(sz) = max_size_nt(nts[i]) else {
+                    return None;
+                };
+                ct += sz;
+                i += 1;
+            }
+            Some(ct)
+        },
+        DataModelType::Map { key, val } => {
```
jamesmunns, line 50: This is wrong, Map is actually basically `Seq((Key, Val))`, this should return None

source/postcard-schema/tests/max_size.rs
```diff
@@ -0,0 +1,28 @@
...
+    a: Result<i32, u32>,
+    b: i128,
+    c: bool,
+}
+
+#[test]
+fn smoke() {
+    let max1 = max_size::<ExampleE>();
+    let max2 = max_size::<ExampleS>();
+    assert_eq!(Some(34), max1);
```
jamesmunns, line 26: This actually overestimates Result a bit - we should count the number of variants and use that to figure out how many varint bits it would be, for example if we only have 2 variants, we only need to add 1 byte for the discriminant, not the full 5.

source/postcard-schema/src/max_size.rs
```diff
@@ -0,0 +1,122 @@
...
+                };
+
+                if sz > max {
+                    max = sz;
+                }
+
+                i += 1;
+            }
+            // discriminants are `varint(u32)`
+            Some(max + 5)
```
jamesmunns, line 116: See comment below, this should be something like `max + min(1, ((nvars + 6) / 7))` or something like that.

source/postcard-schema/src/max_size.rs
```diff
@@ -0,0 +1,124 @@
...
+                    max = sz;
+                }
+
+                i += 1;
+            }
+            let disc_size = if nvars.is_empty() {
+                1
+            } else {
+                // CEIL(variants / 7)
+                (nvars.len() + 6) / 7
```
jamesmunns, line 117: This isn't right, it should be `2^7` not `7`, we want to check how many variants fit in how many varint bytes.

source/postcard-schema/src/lib.rs
```diff
@@ -14,4 +16,25 @@ pub trait Schema {
...
+    /// `N * max_size::<T>`, which is calcuable.
+    ///
+    /// This value should not typically be used directly, instead use
+    /// [`max_size()`][crate::max_size::max_size()], which will EITHER
+    /// use the manual max size, or calculate the max size from the
+    /// schema.
+    ///
+    /// You must not rely on this value for safety reasons, as implementations
+    /// could be wrong.
+    const MANUAL_MAX_SIZE: Option<usize> = None;
```
max-heller, line 39: Instead of having this manual override and a `max_size()` function that checks it, this could default to the calculated size and be overridden as appropriate:

```rust
trait Schema {
    const MAX_SIZE: Option<usize> = max_size_nt(Self::SCHEMA);
}

// Manual max size
impl Schema for heapless::Vec<T, N> {
    const MAX_SIZE: Option<usize> = Some(...);
}
```

This seems more intuitive to me, since there'd be only one source of truth--`T::MAX_SIZE`--instead of the hint and the calculator function.

source/postcard-schema/src/max_size.rs
```diff
@@ -0,0 +1,199 @@
...
+    max_size_dmt(nt.ty)
+}
+
+/// .
+pub const fn bounded_seq_max<Outer: Schema, Inner: Schema, const N: usize>() -> Option<usize> {
+    let DataModelType::Seq(nt) = Outer::SCHEMA.ty else {
+        panic!("Not a bounded seq!");
+    };
+    assert!(nty_eq(nt, Inner::SCHEMA), "Mismatched Outer/Inner types!");
+    let size_one = max_size::<Inner>();
```
max-heller, line 31: Is this dance required to use `Inner::MANUAL_MAX_SIZE` if it's set (as opposed to using `max_size_nt()`)?

Somewhat relatedly, I'm wondering if it would make sense to embed known max sizes in the schema types somewhere in order for `postcard-dyn` to be able to take them into account. Currently, from my understanding, a client using `postcard-dyn` couldn't be told the max size of a `heapless::Vec<T, N>` because the only information in the serialized schema is the unbounded `Seq(T::SCHEMA)`. If the serialized schema included the max length, then that length could be used to size buffers even without knowing the concrete type.

Concretely, this could take the form of bundling an overall max size with an `OwnedNameType`:
```rust
struct OwnedSchema {
    ty: OwnedNamedType,
    max_size: Option<usize>,
}
```
Or it could be more tightly coupled with the schema types themselves, something like:
```rust
struct NamedType {
    name: &'static str,
    ty: &'static DataModelType,
    max_size: Option<usize>,
}
// In this case, Schema could remain as just a single associated const
trait Schema {
    const SCHEMA: &'static schema::NamedType;
}
```

max-heller (commented):
A couple thoughts that have been bouncing around in my head

source/postcard-schema/src/max_size.rs
max-heller: Or, less flexibly, unbounded data model types (strings, seqs, maps, etc.) could have an optional max length field. Then the max size could always be determined based on the schema. Though that might not play as nicely with schema hashing—are two schemas interchangeable if they have different max lengths for a contained string?

max-heller: > Or, less flexibly, unbounded data model types (strings, seqs, maps, etc.) could have an optional max length field. Then the max size could always be determined based on the schema. Though that might not play as nicely with schema hashing—are two schemas interchangeable if they have different max lengths for a contained string?

I'm starting to think this might be the best option:
1. It's harder to implement in subtly wrong ways (e.g., by assuming integers take the same number of bytes as their in-memory representations, which I've seen in the wild)
2. It conveys additional information (bounds on the length of collections) that could be calculated based on the max size, but not as obviously
3. If max size is needed in a `postcard-dyn`/`postcard-rpc` context, it's more compact to send `N` on the wire than `N * max_size(T)`

@jamesmunns can you think of any cases where max size could not be determined this way?

jamesmunns: Just checking in that you were right all along and I'm implementing it this way now :)

PR #188, opened 2024-11-24
Fix nalgebra schema

### LINKED ISSUE #181: postcard-schema: `nalgebra` schema is incorrect
Hey @avsaase - I was wondering how you chose `Seq(T)` for the schema of `nalgebra::Matrix`. A `Seq` is a length prefixed item, so a 3x3 matrix of `u8`s should be 10 bytes. However this test fails:

```rust
use nalgebra_v0_33::{Const, Matrix};
#[test]
fn smoke() {
    let x = nalgebra_v0_33::Matrix::<u8, Const<3>, Const<3>, _>::new(
        1, 2, 3,
        4, 5, 6,
        7, 8, 9,
    );
    let y = postcard::to_stdvec(&x).unwrap();
    assert_eq!(&[9, 1, 4, 7, 2, 5, 8, 3, 6, 9], y.as_slice());
}
```

```txt
test impls::nalgebra_v0_33::test::smoke ... FAILED

failures:

---- impls::nalgebra_v0_33::test::smoke stdout ----
thread 'impls::nalgebra_v0_33::test::smoke' panicked at source/postcard-schema/src/impls/nalgebra_v0_33.rs:32:9:
assertion `left == right` failed
  left: [9, 1, 4, 7, 2, 5, 8, 3, 6, 9]
 right: [1, 4, 7, 2, 5, 8, 3, 6, 9]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Was the choice of `Seq(T)` a guess? Or were you basing this on something else? Just making sure I understand the original choice before I change it.

It appears to me that the actual serialized form is actually `[u8; 9]`, which is *not* a length prefixed, but is actually modeled as a tuple of size N, which avoids serializing a fixed size len:

```rust
impl<T: Schema, const N: usize> Schema for [T; N] {
    const SCHEMA: &'static NamedType = &NamedType {
        name: "[T; N]",
        ty: &DataModelType::Tuple(&[T::SCHEMA; N]),
    };
}
```

max-heller:
Closes #181, using a slice-of-array flattening function borrowed from the const-unstable [`<[[T; N]]>::as_flattened()`](https://doc.rust-lang.org/core/primitive.slice.html#method.as_flattened)

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -29,7 +29,7 @@ cargo_test() {
   cargo test --all "$@"
 }
 
-cargo_test --features=alloc,experimental-derive,use-std,use-crc,derive
+cargo_test --features=alloc,experimental-derive,use-std,use-crc,derive,nalgebra-v0_33
 
 # NOTE: we exclude postcard-dyn for these checks because it is std-only
 
--- a/source/postcard-schema/Cargo.toml
+++ b/source/postcard-schema/Cargo.toml
@@ -68,6 +68,12 @@ path = "../postcard"
 version = "1.0"
 features = ["use-std"]
 
+[dev-dependencies.nalgebra_v0_33]
+package = "nalgebra"
+version = "0.33.0"
+default-features = false
+features = ["serde-serialize-no-std"]
+
 [features]
 default = []
 use-std = ["serde/std"]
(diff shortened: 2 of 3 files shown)


jamesmunns:
Oh this is DEVIOUS. I was ready to come in and say that `&[T]` isn't the same as `&[T; N]`, but I didn't even THINK of doing shenanigans like this *only on the schema*.

I want to poke around at this, but this is extremely clever, thank you!

PR #190, opened 2024-11-27
Make `IOReader` and `EIOReader` constructors public

### LINKED ISSUE #161: Can't create IOReader
I am trying to read from std::io::Read using CRC. As this feature in not standard I am trying to use Flavors.
However,  I can't create an IOReader as its new() constructor is a private associated function, and if I try to crate the struct directly I can't because I need a SlidingBuffer, which is a private struct.
There is a reason why this is the case or am I missing something?

Spartan2909:
Fixes #161

### DIFF
--- a/source/postcard/src/de/flavors.rs
+++ b/source/postcard/src/de/flavors.rs
@@ -261,7 +261,7 @@ pub mod io {
         where
             T: crate::eio::Read,
         {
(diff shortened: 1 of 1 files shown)


jamesmunns:
Thank you for the quick fix @Spartan2909 !

PR #191, opened 2024-11-28
docs: Add notes on serde attributes

ManevilleF:
Add a section in the README warning about the use of serde attributes


jamesmunns:
Would you mind including a link to the "attributes" section of the Serde docs? https://serde.rs/attributes.html

Otherwise, looks reasonable to me.

jamesmunns (approved):
Thank you!

PR #193, opened 2024-11-30
Implement MaxSize for serializable Range structs

ChocolateLoverRaj:
Range, RangeInclusive, RangeFrom, RangeTo


jamesmunns (approved):
Thanks! If you're interested, you could add these to the `builtins_nostd.rs` part of `postcard-schema` as well - I hope to move from a derive macro to a const fn (using the info from postcard-schema) in the future (see #179).

PR #194, opened 2024-12-01
postcard-dyn: reserialize into arbitrary Serializers

max-heller:
Implements a `postcard-dyn` "reserializer"  that deserializes postcard-serialized data and feeds it into an arbitrary `Serializer` in a streaming fashion (without needing to construct an intermediate representation in memory). I think this is (or could be) fully lossless. This means `postcard -> serde_json::Value` (or whatever generic representation `postcard-dyn` ends up providing), `postcard -> json`, and arbitrary other `postcard -> X` transformations can be performed by the same generic implementation.

A few implementation limitations:
- `Serialize` methods require `&'static str`s for structs/variants/field names. To provide these, the current implementation allocates and **leaks** one copy of each unique string that must be used as `&'static`. I can't find a way around this that isn't lossy and this seems inevitable in order to provide a `Serialize`able `postcard_dyn::Value`.
- This is implemented as a `Serialize`able struct that does the deserializing, which is the most straightforward way I found to implement the transformation. However, this means methods only receive `&self`, which necessitates use of `RefCell` for getting exclusive access to the `Flavor`.

@jamesmunns wanted to get your thoughts before I put more work into it. I think it's promising, but the first limitation is... rather limiting.


jamesmunns:
Ouch, yeah, that first limitiation. I don't really love that in the "always" case, do you think this would make sense as an alternate interface in this crate instead of replacing the existing machinery?

max-heller:
> Ouch, yeah, that first limitiation. I don't really love that in the "always" case, do you think this would make sense as an alternate interface in this crate instead of replacing the existing machinery?

Yes, probably, and that should be doable to provide both interfaces backed by a single implementation with a parameter selecting either "lossy, non-leaky transformation" or "lossless, leaky" transformation.

However, what do we expect users to do with a `Value` that doesn't involve `Serialize`ing it in the end (which runs into this limitation)? I'm worried that most transformations on the generic form will end up running into this limitation as well. With the existing interface it isn't a problem because `serde_json` converts structs and enums to maps, but I'm assuming that isn't the end-goal.

max-heller:
One other option: we could expose an `unsafe` interface for serializers that are known not to keep struct/field/variant names around for longer than the serializer lives (I believe this is the case for most serializers) that `transmute`s the lifetimes of those strings to `'static` during serialization.

max-heller:
Pushed an update that avoids leaking for the `from_slice_dyn()` case and makes it easier to implement different workarounds for the `&'static str` limitation. Lots of boilerplate but hopefully reasonably straightforward

max-heller:
Some more context on the `&'static str` limitation and what it may take to lift it in `serde` itself (TLDR: unlikely to ever happen):  https://github.com/serde-rs/serde/issues/2218#issuecomment-2523876811

PR #199, opened 2025-01-16
Add Schema to PathBuf

bugadani:


source/postcard-schema/src/impls/builtins_std.rs
```diff
@@ -21,6 +21,14 @@ impl Schema for std::string::String {
     };
 }
 
+#[cfg_attr(docsrs, doc(cfg(any(feature = "alloc", feature = "use-std"))))]
```
max-heller, line 24: `PathBuf` doesn't exist in `alloc` so this should probably be `#[cfg_attr(docsrs, doc(cfg(feature = "use-std")))]`

max-heller (dismissed):

source/postcard-schema/src/impls/builtins_std.rs
```diff
@@ -21,6 +21,14 @@ impl Schema for std::string::String {
...
+impl Schema for std::path::PathBuf {
+    const SCHEMA: &'static NamedType = &NamedType {
+        name: "PathBuf",
+        ty: &DataModelType::UnitStruct,
```
jamesmunns, line 28: This is probably not correct, do you know how PathBuf actually gets serialized? Is it serialized as a string?

It appears so: https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L957-L980

In general, the `ty` needs to match exactly how the type is actually serialized/deserialized, otherwise libraries like `postcard-dyn` won't work correctly.

jamesmunns (changes requested):

source/postcard-schema/src/impls/builtins_std.rs
jamesmunns: Yeah, @bugadani it would also be good to copy the impl over to the `builtins_alloc` item as well.

source/postcard-schema/src/impls/builtins_std.rs
bugadani: If I say `ty: &DataModelType::String`, `to_pseudocode` will print `String` for `PathBuf`. If that's all right, I'm happy to change it.

jamesmunns: Yeah, the alternative would be to make this a `TupleStruct`, with `String` as the inner type, and "PathBuf" as the name of the outer struct.

I'd have to check if we do that anywhere else for consistency, but I'm less fussed about that as that is more aesthetic than functional.

max-heller: That's to be expected since `Serialize`/`Deserialize` treats `PathBuf` and `String` in the exact same way

jamesmunns: Ah, sorry, probably `NewTypeStruct`, not `TupleStruct`.

max-heller: > Yeah, the alternative would be to make this a `TupleStruct`, with `String` as the inner type, and "PathBuf" as the name of the outer struct.
> 
> I'd have to check if we do that anywhere else for consistency, but I'm less fussed about that as that is more aesthetic than functional.

`NewTypeStruct(T)` vs `T` may be a functional difference as far as `postcard-dyn` is concerned

source/postcard-schema/src/impls/builtins_std.rs
bugadani: Sorry for the noise, I got a bit confused how this all works. I can't get `builtins_alloc` to display locally at all, so I'm not sure what I did is correct.

jamesmunns (approved):
I should probably add snapshot testing/postcard-dyn round tripping, but that's outside this PR. Thank you @bugadani!

PR #202, opened 2025-01-20
Fix EOF not being handled correctly

sommd:
In `IOReader` and `EIOReader`, `pop()` does not properly handle EOF. The `read()` methods return `Ok(0)` when reaching the end of the file:

- https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read
- https://docs.rs/embedded-io/latest/embedded_io/trait.Read.html#tymethod.read

Switched to using `read_exact()`, which returns an `Err` if EOF is reached, which is what the code seems to expect.

### DIFF
--- a/source/postcard/src/de/flavors.rs
+++ b/source/postcard/src/de/flavors.rs
@@ -283,7 +283,7 @@ pub mod io {
             fn pop(&mut self) -> Result<u8> {
                 let mut val = [0; 1];
                 self.reader
-                    .read(&mut val)
+                    .read_exact(&mut val)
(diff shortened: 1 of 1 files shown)


max-heller:
Makes sense, could you add a test covering this?

sommd:
Wasn't sure where you'd want to put integration tests (didn't feel like they'd fit in `loopback`), so I just added some unit tests. If you want integration tests, just let me know where I should put them :slightly_smiling_face:

max-heller (approved):
Looks good, thanks for the contribution!

PR #208, opened 2025-02-01
in to_extend take collection by mutable reference

e00E:
Currently the function takes the collection by value. Taking by mutable reference is better because it is more flexible. Whenever you have a value, you can also pass by mutable reference, but sometimes you only have a mutable reference and cannot pass by value.

This also fixes another problem with the current API: It keeps the collection on error. On success it returns it to the caller, but on error it does not. This makes the caller lose the collection.


e00E:
Yes, it is a breaking change. For a non breaking version we could add a `to_extend_mut` function. Would you like that?

PR #210, opened 2025-02-03
allow to_extend to work with reference to write

e00E:
Currently the function takes the collection by value. Taking by mutable reference is better because it is more flexible. Whenever you have a value, you can also pass by mutable reference, but sometimes you only have a mutable reference and cannot pass by value.

This also fixes another problem with the current API: It keeps the collection on error. On success it returns it to the caller, but on error it does not. This makes the caller lose the collection.

I implement this as a new function rather than by changing the existing to_extend so that this is not a breaking change. When postcard is willing to make breaking changes in the future, then it could decide to remove the existing to_extend.

This is an alternative to #208 .

### DIFF
--- a/source/postcard/src/ser/flavors.rs
+++ b/source/postcard/src/ser/flavors.rs
@@ -250,6 +250,47 @@ where
     }
 }
 
+/// Wrapper over a [`core::iter::Extend<u8>`] that implements the flavor trait
+///
+/// This is like [`ExtendFlavor`], but it takes the writer by reference instead of
+/// by value. This allows you to remain in control of the writer.
+pub struct ExtendRefFlavor<'a, T> {
+    iter: &'a mut T,
+}
+
+impl<'a, T> ExtendRefFlavor<'a, T>
+where
+    T: core::iter::Extend<u8>,
+{
+    /// Create a new [`Self`] flavor from a given [`core::iter::Extend<u8>`]
+    pub fn new(iter: &'a mut T) -> Self {
+        Self { iter }
+    }
+}
+
+impl<T> Flavor for ExtendRefFlavor<'_, T>
+where
+    T: core::iter::Extend<u8>,
+{
+    type Output = ();
+
+    #[inline(always)]
+    fn try_push(&mut self, data: u8) -> Result<()> {
+        self.iter.extend([data]);
+        Ok(())
+    }
+
+    #[inline(always)]
+    fn try_extend(&mut self, b: &[u8]) -> Result<()> {
+        self.iter.extend(b.iter().copied());
+        Ok(())
+    }
+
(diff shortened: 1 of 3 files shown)


e00E:
It's not necessary. The only difference is that on a vector you can call `reserve`, which you can't do through Extend. And the standard library already has special handling making extending vectors with slices faster.

Shatur:
Do you think that compiler replaces `self.iter.extend(b.iter().copied());` with memcpy? :thinking:

e00E:
If self is` Vec<u8>` and the other iter is over `[u8]` then yes.

```rust
fn foo(extend: &mut impl Extend<u8>, iter: impl Iterator<Item = u8>) {
    extend.extend(iter);
}

pub fn with_extend(a: &mut Vec<u8>, b: &[u8]) {
    foo(a, b.iter().copied());
}

pub fn with_extend_from_slice(a: &mut Vec<u8>, b: &[u8]) {
    a.extend_from_slice(b);
}
```

Both of the pub functions will reserve the right amount of space in the vector and then call memcpy. https://godbolt.org/z/81o3hvj58

source/postcard/src/ser/flavors.rs
```diff
@@ -250,6 +250,47 @@ where
     }
 }
 
+/// Wrapper over a [`core::iter::Extend<u8>`] that implements the flavor trait
+///
+/// This is like [`ExtendFlavor`], but it takes the writer by reference instead of
+/// by value. This allows you to remain in control of the writer.
+pub struct ExtendRefFlavor<'a, T> {
```
Shatur, line 257: Maybe use the word "mut" instead? When I see "ref", I usually expect `&`.
```suggestion
pub struct ExtendMutFlavor<'a, T> {
```

source/postcard/src/ser/flavors.rs
```diff
@@ -250,6 +250,47 @@ where
...
+    iter: &'a mut T,
```
Shatur, line 258: Maybe call it `bytes`? `Extend` is not a trait for iterator, it represent a collection that can accept an iterator to extend itself.
```suggestion
    bytes: &'a mut T,
```

Shatur (commented):
Thanks, I copied the implementation into my project, works perfectly :)

I left a couple suggestions.

source/postcard/src/ser/flavors.rs
```diff
@@ -250,6 +250,47 @@ where
...
+    pub fn new(iter: &'a mut T) -> Self {
+        Self { iter }
+    }
+}
+
+impl<T> Flavor for ExtendRefFlavor<'_, T>
+where
+    T: core::iter::Extend<u8>,
+{
+    type Output = ();
```
Shatur, line 275: Maybe return used size for convenience?

PR #211, opened 2025-02-03
Update to heapless v0.8

Gerharddc:
This PR updates the heapless dep from v0.7 to v0.8


jamesmunns:
This would be a breaking change and would require postcard to bump to 2.0.

That is why this has not been done yet.

It would break downstream users code.

Issue #140, opened 2024-04-12
How should the size of `from_io`'s "sliding buffer" be chosen?


smoelius:
My question concerns the second component of `from_io`'s `val` argument`:
https://github.com/jamesmunns/postcard/blob/6338601498b23e962632087c44b19830d579227d/src/de/mod.rs#L85
From what I can tell, it is a "sliding buffer":
- https://github.com/jamesmunns/postcard/blob/6338601498b23e962632087c44b19830d579227d/src/de/mod.rs#L90
- https://github.com/jamesmunns/postcard/blob/6338601498b23e962632087c44b19830d579227d/src/de/flavors.rs#L330-L335

In the `loopback` test, the buffer is 2048 bytes: https://github.com/jamesmunns/postcard/blob/6338601498b23e962632087c44b19830d579227d/tests/loopback.rs#L191-L193

Is this a good choice, generally?

Thank you.

cc: @xgroleau

smoelius:
Thank you, @xgroleau. Your response is very helpful. But should this example then work?
```rust
use postcard::{from_io, to_io};
use serde::de::DeserializeOwned;

fn main() {
    fn sanity(_: &impl DeserializeOwned) {}

    let data = String::from("hello");

    sanity(&data);

    let serialized: ::std::vec::Vec<u8> = vec![];
    let ser = to_io(&data, serialized).unwrap();
    {
        let mut buff = [0; 0]; // <-- Changed 2048 to 0.
        let x = ser.clone();
        let deserialized: String = from_io((x.as_slice(), &mut buff)).unwrap().0;
        assert_eq!(data, deserialized);
    }
}
```
I get:
```
thread 'main' panicked at src/main.rs:16:71:
called `Result::unwrap()` on an `Err` value: DeserializeUnexpectedEnd
```
I'm wondering if there may be a bug here: https://github.com/jamesmunns/postcard/blob/6338601498b23e962632087c44b19830d579227d/src/de/deserializer.rs#L387-L392

xgroleau:
You are right, but it seems strings and probably `deserialize_byte_buf` borrow the data first and the allocate. So in the end, that includes allocation actually, not just borrowed data. I misunderstood and didn't verify with the internals, sorry about that.  Not sure if it is a bug more than just a misunderstanding of the actual behavior. 

 @jamesmunns could validate that if it is the actual desired behavior. If it is, then you need to make sure to have enough space for allocated items and we could probably document it to make sure the behavior is clear.

[...]

PR #212, opened 2025-02-03
Support `std::io::Read` flavor with minimal temporary buffer

Calsign:
As discussed in https://github.com/jamesmunns/postcard/issues/140, currently using the `from_io` or `from_eio` deserialization flavors require passing in a buffer which is large enough to hold almost all of the data, even if the type we are deserializing into is fully owned. This PR implements an approach where the buffer only needs to be large enough to hold the borrowed parts of the resulting type, plus the largest individual piece of data in the input (i.e. the longest string or byte sequence).

The current implementation requires storing more data than is necessary in the buffer because it uses the same implementation for borrowed and owned data. For example: https://github.com/jamesmunns/postcard/blob/9f8bbce2792043fdae1b9823d77377defcb39df9/source/postcard/src/de/deserializer.rs#L385

The borrowed version (`deserialize_str`) needs to return a slice bound by `'de`, so it takes space from the buffer. The owned version (`deserialize_string`) doesn't need `'de`, but it still takes space from the buffer anyway.

In this PR I split `Flavor::try_take_n` into three variants: one which borrows from the input data, one which returns a buffer of a const size, and one which returns a "temporary" slice which is borrowed from the input data but has a lifetime tied to the flavor itself, which in practice means only one temporary slice can be borrowed at a time.

I implemented these extra versions of `try_take_n` for the `from_io` and `from_eio` flavors and then switched to using the new versions in the deserializer implementation where possible.

### DIFF
--- a/source/postcard/src/de/flavors.rs
+++ b/source/postcard/src/de/flavors.rs
@@ -116,8 +116,37 @@ pub trait Flavor<'de>: 'de {
         None
     }
 
-    /// Attempt to take the next `ct` bytes from the serialized message
-    fn try_take_n(&mut self, ct: usize) -> Result<&'de [u8]>;
+    /// Attempt to take the next `ct` bytes from the serialized message.
+    ///
+    /// This variant borrows the data from the input for zero-copy deserialization. If zero-copy
+    /// deserialization is not necessary, prefer to use `try_take_n_temp` instead.
+    fn try_take_n_borrowed(&mut self, ct: usize) -> Result<&'de [u8]>;
+
+    /// Attempt to take the next `N` bytes from the serialized message.
+    ///
+    /// This variant copies the data into a fixed-size buffer.
+    fn try_take_n_const<const N: usize>(&mut self) -> Result<[u8; N]> {
+        Ok(self.try_take_n_borrowed(N)?.try_into().unwrap())
+    }
+
+    /// Attempt to take the next `ct` bytes from the serialized message.
+    ///
+    /// This variant does not guarantee that the returned value is borrowed from the input, so it
+    /// cannot be used for zero-copy deserialization, but it also avoids needing to potentially
+    /// allocate a data in a temporary buffer.
+    ///
+    /// This variant should be used instead of `try_take_n`
+    /// if zero-copy deserialization is not necessary.
+    ///
+    /// It is only necessary to implement this method if the flavor requires storing data in a
+    /// temporary buffer in order to implement the borrow semantics, e.g. the `std::io::Read`
+    /// flavor.
+    fn try_take_n_temp<'a>(&'a mut self, ct: usize) -> Result<&'a [u8]>
+    where
+        'de: 'a,
+    {
+        self.try_take_n_borrowed(ct)
+    }
 
     /// Complete the deserialization process.
     ///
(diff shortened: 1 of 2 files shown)


source/postcard/src/de/flavors.rs
```diff
@@ -116,8 +116,37 @@ pub trait Flavor<'de>: 'de {
...
+    /// Attempt to take the next `ct` bytes from the serialized message.
+    ///
+    /// This variant borrows the data from the input for zero-copy deserialization. If zero-copy
+    /// deserialization is not necessary, prefer to use `try_take_n_temp` instead.
+    fn try_take_n_borrowed(&mut self, ct: usize) -> Result<&'de [u8]>;
+
+    /// Attempt to take the next `N` bytes from the serialized message.
+    ///
+    /// This variant copies the data into a fixed-size buffer.
+    fn try_take_n_const<const N: usize>(&mut self) -> Result<[u8; N]> {
```
jamesmunns, line 128: This could end up returning a fairly large value down the stack, I also worry a bit that it could lead to stack overflow.

Is there a reason you prefer this over the existing try_take_n? There already needs to be bounds checking on the input slice, and I'd be surprised if the compiler didn't inline and optimize based on the calling slice's size.

jamesmunns (commented):
No huge objections, but as-is this would be a breaking change and would have to wait for postcard 2.0. I think it could be possible to handle without a breaking change, if we avoid renaming the existing trait methods, and continue to ensure that all new trait method implementations have default impls. It would be good to check with cargo-semver-check though.

source/postcard/src/de/flavors.rs
```diff
@@ -116,8 +116,37 @@ pub trait Flavor<'de>: 'de {
         None
     }
 
-    /// Attempt to take the next `ct` bytes from the serialized message
-    fn try_take_n(&mut self, ct: usize) -> Result<&'de [u8]>;
+    /// Attempt to take the next `ct` bytes from the serialized message.
+    ///
+    /// This variant borrows the data from the input for zero-copy deserialization. If zero-copy
+    /// deserialization is not necessary, prefer to use `try_take_n_temp` instead.
+    fn try_take_n_borrowed(&mut self, ct: usize) -> Result<&'de [u8]>;
```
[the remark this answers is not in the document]
Calsign: Ah, I forgot that `Flavor` is part of the public API. I wanted to make the distinction between the methods clear in their names. I don't think some combination of deprecation and a default impl would improve much here, so I'll just change the name back to `try_take_n`.

source/postcard/src/de/flavors.rs
Calsign: Good point. I'll switch to using `try_take_n_temp`.

Calsign:
OK, I've addressed your comments. `try_take_n` isn't renamed anymore, and `try_take_n_temp` has a default impl.

I ran cargo-semver-checks and it doesn't report any issues for `postcard` (but there might be issues in `postcard-schema` unrelated to this PR). I needed to comment [these lines](https://github.com/jamesmunns/postcard/blob/9f8bbce2792043fdae1b9823d77377defcb39df9/source/postcard/src/eio.rs#L3-L10) to get it to work (I think semver-checks is passing `--all-features`), I'm not sure if there's a preferred way to do that.

jamesmunns:
Thank you for resolving the semver issues!

Issue #214, opened 2025-02-07
Consider implementing `Serialize, Deserialize` traits for `postcard::Error`


quietlychris:
I'm working on doing some error propagation where `postcard::from_bytes()` methods might return a `postcard:Error`, and am trying to encapsulate that into a larger `MyError` type that currently implements serde's `Serialize, Deserialize` traits (ironically enough, so that I could potentially serialize that error and send it across the wire). 

I know it's technically possible to wrap this in my own library using serde's remote derive functionality or using `map_err(|e| e.into())` for my own wrapper type, but it would more convenient if `postcard::Error` were to implement those traits on their own. Is this something that you might be willing to consider?

jamesmunns:
I think a PR for this would be good! I can't think of any reason offhand that would be a problem.

PR #215, opened 2025-02-07
Add Serialize and Deserialize traits to postcard::Error

quietlychris:
This pull request addresses #214. As far as I can tell was a one-line change since both traits were already available, but if there's any issues, just let me know :)


jamesmunns:
Looks reasonable to me! If CI passes, I'll hit merge and it'll be in the next release.

jamesmunns:
Thanks!

PR #218, opened 2025-03-06
Fix typo in docstring

lecanard539:
The docs for `FeedResult::DeserError` say 

> Contains remaining section of input, if. any 

but it should be 

> Contains remaining section of input, if any.

### DIFF
--- a/source/postcard/src/accumulator.rs
+++ b/source/postcard/src/accumulator.rs
@@ -76,8 +76,7 @@ pub enum FeedResult<'a, T> {
     /// Buffer was filled. Contains remaining section of input, if any.
     OverFull(&'a [u8]),
 
-    /// Reached end of chunk, but deserialization failed. Contains remaining section of input, if.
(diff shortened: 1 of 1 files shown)


source/postcard/src/accumulator.rs
```diff
@@ -76,8 +76,7 @@ pub enum FeedResult<'a, T> {
     /// Buffer was filled. Contains remaining section of input, if any.
     OverFull(&'a [u8]),
 
-    /// Reached end of chunk, but deserialization failed. Contains remaining section of input, if.
-    /// any
+    /// Reached end of chunk, but deserialization failed. Contains remaining section of input, if any
```
max-heller, line 79: For consistency with the other variants, let's add a period to the end

max-heller (changes requested):
Thanks, one tiny nit

max-heller (approved):

PR #220, opened 2025-04-06
Add initial implementation of Postcard-C

sphw:
This PR adds a single header file implementation of the postcard spec in C. It's designed to be a relatively low-level API that allows users to manually write serialization and deserialization logic. It uses a BYOB (bring your own buffer) design, where the user has to pass a pre-allocated buffer of data. As such, it does not do any memory management on its own. 

You can find more details on usage in [`README.md`](https://github.com/sphw/postcard/blob/e1266a88e4a168d55dd9867ea3749fc4b3f67672/source/postcard-c/README.md) and [`example.c`](https://github.com/sphw/postcard/blob/e1266a88e4a168d55dd9867ea3749fc4b3f67672/source/postcard-c/example.c)

I've tested this implementation as part of a C++ code generation system, and it works well on the types I've thrown at it. That being said, I can't guarantee that it is bug-free, and I would like to add a test suite that tests this implementation against the Rust implementation. 

I will also be open-sourcing a C++ code generator that uses postcard-schema and postcard-c to generate high-level C++ structs.

### DIFF
--- a/source/postcard-c/example.c
+++ b/source/postcard-c/example.c
@@ -1,0 +1,121 @@
+#include "postcard.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+/// helper function that prints out a byte buffer to stdout
+void print_buffer(uint8_t* buffer, size_t len)
+{
+    printf("serialized data [");
+    for (size_t i = 0; i < len; i++) {
+        printf("%d", buffer[i]);
+        if (i < len - 1) {
+            printf(", ");
+        }
+    }
+    printf("]\n");
+}
+
+/// helper function that prints out an array of int16 values to stdout
+void print_values(int16_t* values, size_t len)
+{
+    printf("values: [");
+    for (size_t i = 0; i < len; i++) {
+        printf("%d", values[i]);
+        if (i < len - 1) {
+            printf(", ");
+        }
+    }
+    printf("]\n");
+}
+
+int main()
+{
+
+    // This example shows manually serializing and deserializing the Foo struct below.
+    //
+    // ```rust
+    // struct Foo {
+    //     id: u32,
+    //     name: String,
+    //     values: Vec<i16> // len 3
+    //     is_active: bool
+    // }
+    // ```
+
+    // allocate a buffer large enough to fit the serialized data
+    uint8_t buffer[128];
+
+    // create a new `postcard_slice_t` that uses `buffer` as the underlying storage
+    // `postcard_slice_t` is a growable reference to some underlying buffer.
+    postcard_slice_t slice;
+    postcard_init_slice(&slice, buffer, sizeof(buffer));
+
+    // encode id
+    postcard_encode_u32(&slice, 1234);
+
+    // encode name
+    const char* name = "PostcardTest";
+    postcard_encode_string(&slice, name, strlen(name));
+
+    // encode the 3 values
+    postcard_start_seq(&slice, 3);
+    postcard_encode_i16(&slice, -10);
+    postcard_encode_i16(&slice, 20);
+    postcard_encode_i16(&slice, -30);
+
+    // encode is_active
+    postcard_encode_bool(&slice, true);
+
+    // print the encoded data
+    // slice.len now contains the length of the serialized
+    // data from the serialization function
+    print_buffer(slice.data, slice.len);
+
+    // to decode the data we will create a new slice
+    // in the decode path, `postcard_slice_t.len` is used
+    // as a cursor for the decoded data. So we need a new slice
+    // that will only reference the serialized data, and
+    // have len = 0
+    postcard_slice_t decode_slice;
+    postcard_init_slice(&decode_slice, buffer, slice.len);
+
+    // decode id
+    uint32_t id;
+    postcard_decode_u32(&decode_slice, &id);
+    printf("id: %u\n", id);
+
+    // decode name
+    size_t actual_len;
+    // first we decode the length of the string
+    // this is pulled out as a different function,
+    // so you can heap-allocate a string based on the actual
+    // length of the string
+    postcard_decode_string_len(&decode_slice, &actual_len);
+    char name_buffer[actual_len + 1];
+    postcard_decode_string(&decode_slice, name_buffer, sizeof(name_buffer),
+        actual_len);
+    name_buffer[actual_len] = '\0'; // null terminate the string
+    // strings in postcard are encoded as byte arrays of valid utf8 data,
+    // and then do not include a null terminator. We add this terminator
+    // manually so we can print the string using `printf`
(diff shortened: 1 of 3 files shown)


source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+/// The user must ensure that bytes points to a valid initialized memory, and
+/// that length is not greater than the size of the array.
+///
+/// *Side Effects /
+/// Returns* If there is not enough room in the buffer `postcard_error_t` will
+/// return a non-zero value If encoding was successful, slice.len will be
+/// incremented by the number of encoded bytes
+postcard_error_t postcard_encode_byte_array(postcard_slice_t* slice,
+    const uint8_t* bytes,
+    size_t length);
```
i509VCB, line 182: Would it be helpful to have a way to determine if there is enough room in the slice to push the array, without actually writing anything? (And other functions like this)

source/postcard-c/example.c
```diff
@@ -0,0 +1,121 @@
+#include "postcard.h"
```
i509VCB, line 1: There should probably be an SPDX header here. I do plan to use Postcard-C in projects where this is required.

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+} postcard_slice_t;
+
+/// postcard_c error codes
+typedef enum {
+    POSTCARD_SUCCESS = 0,
+    POSTCARD_ERROR_BUFFER_TOO_SMALL,
+    POSTCARD_ERROR_INVALID_INPUT,
+    POSTCARD_ERROR_INCOMPLETE_DATA,
+    POSTCARD_ERROR_OVERFLOW
+} postcard_error_t;
```
i509VCB, line 26: I would prefer each error code to get an explicitly defined value. Whether the error values are positive or negative values is a different question.

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+postcard_error_t postcard_encode_byte_array(postcard_slice_t* slice,
+    const uint8_t* bytes,
+    size_t length);
+
+/// Encodes a string to the passed in slice.
+///
+/// *Arguments*
+/// - slice - a pointer to an initialized `postcard_slice_t`
+/// - string - a pointer to the string to encode. This string should be utf8
+/// encoded string. It shouldn't be null-terminated.
```
i509VCB, line 189: Is `should` a strong enough guarantee here? Or should it be considered undefined behavior if the string is NOT UTF-8?

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+/// - length - the length of the string
```
i509VCB, line 190: To make it more clear that this is talking about each byte to be written and not the number of codepoints.

```suggestion
/// - length - the length of the string in bytes
```

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
+#ifndef POSTCARD_H
```
i509VCB, line 1: Top of header documentation should probably include a link to the specification.

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+inline postcard_error_t postcard_encode_f32(postcard_slice_t* slice,
+    float value)
+{
+    if (!slice || !slice->data)
+        return POSTCARD_ERROR_INVALID_INPUT;
+    if (slice->len + 4 > slice->capacity)
+        return POSTCARD_ERROR_BUFFER_TOO_SMALL;
+
+    uint32_t bits;
+    memcpy(&bits, &value, sizeof(bits));
```
i509VCB, line 712: Should we use the memcpy in the standard library or make the caller pick the memcpy?

One pattern I often see is that a library has a seperate memcpy function, like a `postcard_memcpy`. This would let the caller define it's own memcpy implementation.

i509VCB (commented):
I also noticed that encoding `u128` and `i128` is not currently possible. Although I do understand that 128-bit types are special and require checking the compiler with proprocessor code.

source/postcard-c/postcard.h
sphw: This is a potential problem across all of the functions, not just byte array. The trouble is that it would make encoding essentially a two-step operation: one to check how much size a varint takes, and another to actually encode it. 

In the use cases I envisioned for this, I don't think this will cause practical problems. The first is a higher-level code generator that will pre-allocate memory based on the calculated encoded size. Another is serializing messages to fixed-size buffers. In that case, I'm not too concerned if there is half-written data, as long as the functions properly error out.

I think we can recommend that users run the postcard_size functions themselves if they never want to write data when the buffer isn't big enough.

source/postcard-c/example.c
sphw: Yeah that makes sense. I can add a SPDX header for MIT + Apache dual license to match the rest of the repo

source/postcard-c/postcard.h
sphw: I've been going back and forth on what the best behavior is here. Personally I like to use a minimal libc (i.e newlib) on embedded devices, but I can understand not wanting to rely on that. I think the ergonomic benefits for non-embedded development are important though. 

A good way to do it might be to use `__has_include` to check if `string` is available and use that version of memcpy if it exists. Alternatively we could use a preprocessor definition to allow the user to optionally override the memcpy impl

source/postcard-c/postcard.h
rursprung: a bit tangential to the discussion, but relevant for this (and similar APIs):
rather than passing a pointer you could pass an array - there's a fancy C notation (since C99) which allows you to specify the parameter which gives the length and the compiler will validate it. but the length parameter has to come before the array, otherwise it doesn't work (until [this proposal](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3394.pdf) gets standardised).

this works:
```c
int foo(unsigned int len, char foo[static len]);
```

source/postcard-c/postcard.h
```diff
@@ -0,0 +1,1184 @@
...
+    // Pointer to the underlying data buffer
+    uint8_t* data;
+    /// The total written length of data in the buffer
+    size_t len; // Current length of data
+    /// The maximum capacity of the buffer
+    size_t capacity;
+} postcard_slice_t;
+
+/// postcard_c error codes
+typedef enum {
```
rursprung, line 20: i know that this is currently pure C, but doesn't pulling it into a C++ code base pollute the namespace (unless you `#include` it within a `namespace`)?
would it make sense to make the header C & C++ and use `#ifdef __cplusplus` copiously to decide between the two? that'd e.g. allow using `enum class` rather than `typedef enum`

PR #221, opened 2025-04-13
Add `postcard-schema` implementations for types from `fixed`

i509VCB:
This are just spicy primitive numbers, so it should be simple enough.


jamesmunns:
Thank you!

PR #222, opened 2025-04-15
Migrate from paste to with_builtin_macros

Gerharddc:
This is to deal with security warnings related to paste now being "unmaintained". These warnings affect everyone using automated security scanners and the "crc" feature.

Please see https://rustsec.org/advisories/RUSTSEC-2024-0436.html, https://github.com/rustsec/advisory-db/issues/2203 and https://github.com/rustsec/advisory-db/pull/2215 for more details.

### DIFF
--- a/source/postcard/Cargo.toml
+++ b/source/postcard/Cargo.toml
@@ -61,8 +61,8 @@ optional = true
 version = "3.0.1"
 optional = true
 
-[dependencies.paste]
-version = "1.0.12"
+[dependencies.with_builtin_macros]
+version = "0.1.0"
 optional = true
 
 [features]
@@ -80,7 +80,7 @@ use-std = ["serde/std", "alloc"]
 heapless-cas = ["heapless", "dep:heapless", "heapless/cas"]
 alloc = ["serde/alloc", "embedded-io-04?/alloc", "embedded-io-06?/alloc"]
 use-defmt = ["defmt"]
-use-crc = ["crc", "paste"]
+use-crc = ["crc", "with_builtin_macros"]
 
 # Experimental features!
 #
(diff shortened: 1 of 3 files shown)


jamesmunns:
Thank you for the heads up and the PR @Gerharddc, although it is somewhat disappointing to me to switch from a 1.0 (but unmaintained) crate from dtolnay (a well known author), to a 0.1 dep from someone I don't know. I feel like the risk of a vuln being found in our simple use of `paste` is pretty low, and the chance of `paste` being supply-chain-attacked is also decently low.

Honestly, since it is only used for 5 invocations, I'd rather just manually copy and paste the impls and remove paste/with_builtin_macros entirely, if removing paste is important to folks.

Gerharddc:
@jamesmunns understood. I'm not crazy about the idea of copy and pasting such large/complex-ish blocks of code 5 times for each of the 5 flavors though. To me that would look a bit too ugly and be harder to maintain.

Perhaps a middle ground could be to modify the `impl_flavor` macro so that the function idents get passed in instead of being generated. Then

`impl_flavor![u8, u16, u32, u64, u128];`

would become something like

```
impl_flavor![u8, from_bytes_u8, take_from_bytes_u8];
impl_flavor![u16, from_bytes_u16, take_from_bytes_u16];
...
```

What do you think about that solution?

jamesmunns:
That works for me too!

Gerharddc:
Ok, I have now implemented the proposed change and it seems to work from what I can tell.

jamesmunns (approved):

jamesmunns:
@Gerharddc minor formatting nit in CI:

```
error: doc list item overindented
 --> source/postcard/src/de/flavors.rs:8:5
  |
8 | //!      appended to the message, etc.
  |     ^^^^^ help: try using `   ` (3 spaces)
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_overindented_list_items
  = note: `-D clippy::doc-overindented-list-items` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(clippy::doc_overindented_list_items)]`
```

Gerharddc:
@jamesmunns thanks! seems I missed it locally because I had a slightly outdated clippy

PR #227, opened 2025-05-16
Fix missing license files in published crates

musicinmybrain:
Add symlinks to the top-level license files to fix missing license files in published crates.

Before:

```
[…]/source/postcard $ cargo publish --dry-run
[…]/source/postcard $  ls ../../target/package/postcard-1.1.1
Cargo.lock  Cargo.toml  Cargo.toml.orig  README.md  src  target  tests
```

After:

```
[…]/source/postcard $ cargo publish --dry-run
[…]/source/postcard $  ls ../../target/package/postcard-1.1.1
Cargo.lock  Cargo.toml  Cargo.toml.orig  LICENSE-APACHE  LICENSE-MIT  README.md  src  target  tests
```

Note that both chosen licenses require their text to be distributed.

### DIFF
--- a/source/postcard-derive/LICENSE-APACHE
+++ b/source/postcard-derive/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../../LICENSE-APACHE
--- a/source/postcard-derive/LICENSE-MIT
+++ b/source/postcard-derive/LICENSE-MIT
@@ -1,0 +1,1 @@
+../../LICENSE-MIT
--- a/source/postcard-dyn/LICENSE-APACHE
+++ b/source/postcard-dyn/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../../LICENSE-APACHE
--- a/source/postcard-dyn/LICENSE-MIT
+++ b/source/postcard-dyn/LICENSE-MIT
@@ -1,0 +1,1 @@
+../../LICENSE-MIT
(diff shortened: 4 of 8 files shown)


jamesmunns:
Grrr, I missed adding this PR to my triage list, I'll add it now. Thanks for the reminder.

musicinmybrain:
> Just a quick question @musicinmybrain - do you know if symlinks like this will work well on windows/mac/linux? I know there is some subtlety wrt symlinks/hardlinks, but I'm not sure. If you aren't either, I'll just fully copy the files into all of the subrepos, or figure out how to instruct the packaging to also include those files.

It works fine on Linux and Mac. On Windows, I believe the situation is more subtle. It’s fine on [WSL](https://learn.microsoft.com/en-us/windows/wsl/); otherwise it requires some configuration, https://github.com/o2sh/onefetch/issues/1363#issuecomment-2225977013. If you were to `cargo publish` from Windows, not using WSL, without proper configuration, then the symlinks would end up published as regular files containing their target paths (like `../../LICENSE-APACHE`) rather than the contents of the target files. Obviously, that would defeat the purpose of this PR, so if there’s a chance of that happening, then it’s better just to copy the license files.

Symlinking license files like this *is* a very common pattern; many projects just don’t publish from Windows without using WSL. But copying the license files is fine, too.

musicinmybrain:
(You can always test this on a particular platform you have access to by using `cargo publish --dry-run`.)

jamesmunns:
> (You can always test this on a particular platform you have access to by using `cargo publish --dry-run`.)

Yeah, I don't have a Windows box handy :)

PR #230, opened 2025-05-23
Implement Schema for `alloc::boxed::Box<T>`.

rosslannen:
Implements schema for boxed types.

I based the implementation off of the reference implementation, which just delegates to the underlying type. If there is another way that this should be done I can change it.


max-heller (changes requested):
This looks good, although we should add it to `source/postcard-schema/src/impls/builtins_std.rs` as well. (I'm not entirely sure why `builtins_std.rs` contains redundant impls for `alloc`, but it seems that `builtins_alloc.rs` is only compiled when the `std` feature is disabled)

max-heller (approved):
Looks good!

PR #234, opened 2025-06-08
Version bump defmt to 1.x.x

IamTheCarl:
Bumps optional defmt dependency up to version 1.x.x.

I don't know what else to say. This is the smallest PR I've made in my life.


rursprung (approved):
due to the semver trick employed by `defmt` this should be perfectly fine

PR #237, opened 2025-06-13
Reduce unsafety and add safety comments to remaining unsafe code

Veykril:
This brings the MSRV to 1.48 if it is less than that.


jamesmunns:
TIL "as_ptr_range"! Thanks! MSRV shouldn't be a problem, I don't think I promise it anywhere, and 1.48 is sufficiently old enough anyway.

PR #240, opened 2025-06-17
Implement Schema and MaxSize for more core types

max-heller:
- `core::time::Duration`
- `core::marker::PhantomData`
- `core::num::{Wrapping,Saturating}`
  - Added feature flags for `Saturating` since it's relatively new (1.74)

### DIFF
--- a/source/postcard/src/max_size.rs
+++ b/source/postcard/src/max_size.rs
@@ -15,6 +15,7 @@ use core::{
         NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize,
     },
     ops::{Range, RangeFrom, RangeInclusive, RangeTo},
+    time::Duration,
 };
 
 /// This trait is used to enforce the maximum size required to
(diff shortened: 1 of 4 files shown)


source/postcard/src/max_size.rs
```diff
@@ -218,6 +223,16 @@ impl<T: MaxSize> MaxSize for RangeTo<T> {
     const POSTCARD_MAX_SIZE: usize = T::POSTCARD_MAX_SIZE;
 }
 
+impl<T: MaxSize> MaxSize for core::num::Wrapping<T> {
+    const POSTCARD_MAX_SIZE: usize = T::POSTCARD_MAX_SIZE;
+}
+
+#[cfg(feature = "experimental-derive-core-num-saturating")]
+#[cfg_attr(docsrs, doc(cfg(feature = "experimental-derive-core-num-saturating")))]
```
jamesmunns, line 231: Could we remove the `experimental-derive-core-num-saturating` feature, and change this to `all(feature = "core-num-saturating", feature = "experimental-derive")` instead?

jamesmunns: I'm gunna go ahead and make this change, shout if you hate it :)

max-heller: Done, left the `#[doc(cfg)]` with just `feature = "core-num-saturating"` because this module is only exported under `feature = "experimental-derive"`

PR #245, opened 2025-06-20
MaxSize for postcard::Error

### LINKED ISSUE #224: Implement MaxSize for postcard::Error
For similar reasons as #214, it would be useful if postcard::Error had a MaxSize

Would this be possible for 2.0?

cramt:
fixes #224

### DIFF
--- a/source/postcard/src/error.rs
+++ b/source/postcard/src/error.rs
@@ -1,6 +1,8 @@
 use core::fmt::{Display, Formatter};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


jamesmunns (changes requested):
I think this is correct currently, but if we ever add an error variant that isn't just a discriminant, this could be subtly wrong. Also, `varint_max` is mostly intended for integers directly.

Would it be possible to use the `postcard-derive` `MaxSize` derive for this instead?

PR #246, opened 2025-06-20
automated insta snapshot testing for postcard-schema

### LINKED ISSUE #204: Consider optional "snapshot" feature for postcard-schema
The derive for Schema could potentially emit `insta` tests that describe the schema of the type, meaning that `icd` crates could automatically gain regression testing abilities.

We'd probably want to gate these behind something like `#![cfg(all(feature = "snapshot", feature = "use-std", test))]`, and add docs to show users how to run these (including first run where they will need to approve the snaps).

cramt:
fixes #204

had to make it skip any generic structs since that wouldnt really make sense. sadly this pr also includes all the snapshot tests now, which is a lot of bloat

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -29,7 +29,7 @@ cargo_test() {
   cargo test --all "$@"
 }
 
-cargo_test --features=alloc,experimental-derive,use-std,use-crc,derive,nalgebra-v0_33,heapless-v0_8
+cargo_test --features=alloc,experimental-derive,use-std,use-crc,derive,nalgebra-v0_33,heapless-v0_8,insta
 
 # NOTE: we exclude postcard-dyn for these checks because it is std-only
 
--- a/source/postcard-derive/Cargo.toml
+++ b/source/postcard-derive/Cargo.toml
@@ -31,3 +31,12 @@ all-features = true
 syn = "2.0"
 quote = "1.0"
 proc-macro2 = "1.0"
+
+[dependencies.insta]
+package = "insta"
+version = "1"
+default-features = false
+optional = true
+
+[features]
+insta = ["dep:insta"]
--- a/source/postcard-derive/src/schema.rs
+++ b/source/postcard-derive/src/schema.rs
@@ -1,8 +1,7 @@
 use proc_macro2::{Span, TokenStream};
-use quote::{quote, quote_spanned};
+use quote::{quote, quote_spanned, ToTokens};
 use syn::{
-    parse_quote, punctuated::Punctuated, spanned::Spanned, Data, DeriveInput, Fields, GenericParam,
-    Generics, Path, Token,
+    parse_quote, punctuated::Punctuated, spanned::Spanned, Data, DeriveInput, Fields, GenericParam, Generics, Ident, Path, Token
 };
 
 pub fn do_derive_schema(input: DeriveInput) -> syn::Result<TokenStream> {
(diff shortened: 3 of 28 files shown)


jamesmunns:
Interesting! I hadn't considered the limitation on generics. I suppose the only other option would be to instead generate a function that could be called in a test context, but isn't a top level "test" itself.

On the upside: you could support generics, because it could be done on each instantiation. On the downside: users don't just get the check for free.

I wonder if for now we should generate a *failing* test for generic types, so we don't mislead/confuse folks? I'm open to your thoughts here!

Also, once we answer the stuff above, it would be good to port this to `postcard-schema-ng` as well.

Thank you for working on this!

jamesmunns:
Yeah, my only caveat is "if one of your deps enabled the insta feature, but you want generics, then you're kind of hosed unless you patch your dep".

It would be nice to have something like `#[postcard(snapshot = skip)]` or so for a type (not a field), like:

```rust
#[derive(Schema)]
#[postcard(snapshot = skip)]
struct Foo<T: Schema> {
     x: T,
     y: usize,
}
```

or maybe just a second derive, like:

```rust
#[derive(SchemaNoSnapshot)]
struct Foo<T: Schema> {
     x: T,
     y: usize,
}
```

But that's kinda weird, and you might still get in the case where one dep has generics and one dep enables insta.

Or maybe just an env var like `POSTCARD_SCHEMA_SKIP_GENERIC_SNAPSHOTS` that make the tests pass/skip?

I don't have a great way to solve this :/. I appreciate you working on this, and I don't want you to get discouraged! I have not thought this all the way through, so we definitely still need to think around the edge cases of this a bit so this doesn't upset any users.

jamesmunns:
Maybe the right choice is having a totally separate derive for snapshot testing to make it really explicitly opt-in? e.g. `#[derive(SchemaSnapshot)]`?

max-heller:
> Though imo doing `#[postcard(snapshot)]` would probably be better than a second derive because then you can accidentally make something derive `SchemaSnapshot` without deriving `Schema`. And then just ending with a compile error if you try to do `#[postcard(snapshot)]` on something with a generic bound more specific than `Schema`

Agreed that an opt-in `#[postcard(snapshot)]` makes sense, it might not be desirable to generate tests for all types implementing `Schema` even if `#[derive(Schema)]` could.

For generic types, what about something like `#[postcard(snapshot(TYPE))]`?
```rust
#[derive(Schema)]
#[postcard(snapshot(Foo<()>))]
struct Foo<T: Schema> {
     x: T,
     y: usize,
}
```

jamesmunns:
I think shipping a first version of this (like `postcard(snapshot)`) that doesn't support generics is fine, as long as we could possibly support it in the future.

I would think in most cased (randomly guessing tho) we'd more often have generic types like:

```rust
// in some library/dep somewhere...
#[derive(Schema)]
struct Foo<T: Schema> {
     x: T,
     y: usize,
}

// actual "leaf" type
#[derive(Schema)]
#[postcard(snapshot)]
struct ActualType {
    foo: Foo<u64>,
}
```

Idk tho! I'm not sure how easy it would be to do:

```rust
#[postcard(snapshot)]
type MyFoo = Foo<u64>;
```

That probably takes us out of derive territory and into general proc macros? Not my strong field 😅

jamesmunns:
Agreed that if we don't make it automatic, we don't need it to be a feature (unless we pull in a bunch more deps or something to support it).

Issue #252, opened 2025-07-07
[postcard-schema] Adding support for uom 0.37 in impl/


PedromcaMartins:
I was trying to use postcard to send Uom (https://crates.io/crates/uom) and nalgebra structs over the wire, but it seems that postcard-schema doesn't implement a schema for uom. 
I'll go ahead, and fork the project to implement this feature. 
I greatly appreciate your feedback, as this is my first Issue ever. 
And also, let me know if you're also interested in creating a schema for uom. 

Cheers ;D

PR #253, opened 2025-07-07
Add impls for uom_v0_37

PedromcaMartins:
PR to implement issue #252. 

Thought process: 
The uom crate is built upon the struct Quantity, with the underlying data representation (/inner type) as a generic V (f32, u8, ...). 
Everything else in the crate is a type alias to a Quantity struct, making the Schema implementation easier. 
(This is kinda similar to the Matrix struct for the nalgebra crate)

In order to implement the Schema trait, I added the Schema trait bound to the inner type V, and just used the schema implementation of the inner type. 
In practive, this means that if I create a variable containing the Acceleration, stored in a float, it will use the schema of the float. 

I'm going to be honest, I don't fully understand what the Schema does under the hood, and if I can just use the schema of the inner type, instead of (for example) making the schema a NewtypeVariant, containing the inner type. 
I'd love some feedback, especially because I'm not that experienced with this codebase. 

Summary: 
- Added uom as dependency + dev-dependency for simple tests. 
- Implemented the Schema trait by adding the trait bound for the inner type. 
- Tested Schema approach comparing with the inner type, and also one conversion test, similar to nalgebra.
- Build Tested in external project, and the compiler was happy :))


jamesmunns:
Thank you for the PR!

I think this implementation is reasonable, re: your question:

> I'm going to be honest, I don't fully understand what the Schema does under the hood, and if I can just use the schema of the inner type, instead of (for example) making the schema a NewtypeVariant, containing the inner type.

This is a bit of a "vibes" question I'm not sure about.

Right now, you've set the schema to be the schema of the inner type. This means that regardless of the "unit", the schema will be the same. For example, `f32::Acceleration` and `f32::ThermodynamicTemperature` will have the same schema.

On one hand: that's right, you are sending an `f32`, and this is how the data is encoded on the wire.

On the other hand, if people are using the schema, or a hash of the schema (as is done in postcard-rpc), people might want different "units" to exist as different types. You could achieve this by making a newtype wrapper with the name of the unit as the type, for example something like `Acceleration(f32)`. Doing this means that in postcard-rpc if you have an endpoint that is supposed to take an `f32::Acceleration` and you send it a `f32::ThermodynamicTemperature`, it will be able to say "no that's wrong".

I don't think there is an objective answer here, but like I said, it's more of a "vibes" question.

CC @max-heller for thoughts.

Also, two requests:

* The CI is failing due to needing a `cargo fmt`
* (before we merge, after we answer the question above) Could you also please port the implementation to the `postcard-schema-ng` crate? That way the next version of postcard-schema will have support for `uom` as well.

PedromcaMartins:
Thanks for all the feedback! 

I understand both sides of the argument. 

It does make sense why they just serialize the inner type. 
They store the quantity (length, time, ...) as phantom data, and after glancing over the serde API, it seems like the type T is lost when serializing and deserializing generic phantom data types... 

I wasn't aware of postcard-dyn, or that it could cause problems with this implementation... 

Since the uom crate uses the type system to enforce two different quantities can't be the same, I think this property should be preserved even after serializing/ deserializing... 
Additionally, I couldn't find any discussion on uom repo about the serialization of the phantom data in order to preserve the quantity before and after serializing... 
I guess it wasn't a concern, but I could create an issue there just for clarification and advice. What are your thoughts on this? 

About the 2 extra requests: 
- I will amend the current commit with `cargo fmt` (I didn't know this was a thing). 
- I'll add the final implementation to postcard-schema-ng. By the way, what does ng stand for?

CBJamo:
Unless it makes postcard-dyn impossible, I have a strong preference for the schema to include the "unit" type, because this:

> Doing this means that in postcard-rpc if you have an endpoint that is supposed to take an `f32::Acceleration` and you send it a `f32::ThermodynamicTemperature`, it will be able to say "no that's wrong".

Is the behavior I (and I think most other users) would expect. If the schema doesn't differentiate between different units, why bother serializing the uom type at all?

max-heller:
> I wonder if we could define the serialization form as something like:
> 
> ```rust
> struct {
>     data: f32,
>     unit: PhantomData<Acceleration>,
> }
> ```

We could do something similar if we defined a wrapper type around `uom::Quantity`, or a [`#[serde(with = "...")]`](https://serde.rs/field-attrs.html#with) style adapter. We can't use this as the schema for `Quantity` as-is because it does not match `Quantity`'s `Serialize`/`Deserialize` impls.

> On the wire, it would still just be `f32` (as zsts are not serialized), and maybe we could teach postcard-dyn to deserialize PhantomData types as the name of the type? e.g. you'd get:
> 
> ```json
> {
>   "data": 1.234,
>   "unit": "Acceleration"
> }
> ```

These statements seem to conflict: there's either just an `f32`, or an `f32` and some unit (string, int, etc.).

> maybe we could teach postcard-dyn to deserialize PhantomData types as the name of the type?

Doing this in the general case would be difficult since [`type_name()`](https://doc.rust-lang.org/std/any/fn.type_name.html) is noisy and unstable. For `uom` in particular, it'd probably make more sense to rely on the `Unit` trait and have custom `Serialize`/`Deserialize` impls that serialize and validate the unit identifier, respectively.

max-heller:
> the intent of my last comment was that adding or removing a phantomdata from the schema doesn't affect what hits the wire, even if it's a semantically different schema.

In postcard it doesn't, but I think it could in a different format given `PhantomData`'s `Serialize` impl [serializes a unit struct](https://docs.rs/serde/1.0.219/src/serde/ser/impls.rs.html#117-128), which in some serde formats may impact wire format. Additionally, in self-describing formats, the change from `10` to `{ data: 10, unit: ... }` will impact wire format because of adding field identifiers.

> I forget if we came to a conclusive answer to "what is 'right' for a schema to contain", but my personal thought is that I care more about _what is on the wire_ than _what serde methods were called_.

My general concern is that this approach will break `postcard-dyn` (especially https://github.com/jamesmunns/postcard/pull/194 if it ends up going anywhere). E.g., converting from JSON with a schema like the one described above would attempt to deserialize an object instead of a literal, when the JSON serialized form of a quantity is just a literal.

PR #269, opened 2025-09-13
Add heapless v0.8 and v0.9 MaxSize impls for postcard 1.x

guineawheek:
While I understand bumping the main `heapless` in this crate is something reserved for postcard 2 as it is a breaking change, it would be nice to optionally have `MaxSize` implementations for newer versions of `heapless` as a stopgap before an upgrade to `postcard` 2 is made at some point in the future, potentially also bringing further refactors that eliminate the need for these impls.

This doesn't touch `to_vec` or similar methods; I don't see a way of editing those without breaking semver and you still get a lot of utility out of `MaxSize` impls anyway.

### DIFF
--- a/source/postcard/Cargo.toml
+++ b/source/postcard/Cargo.toml
@@ -30,6 +30,20 @@ default-features = false
 features = ["serde"]
 optional = true
 
+[dependencies.heapless-v0_8]
+package = "heapless"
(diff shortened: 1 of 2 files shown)


jamesmunns:
@guineawheek slight format nit, if you get this updated I should be able to get this merged.

PR #277, opened 2025-10-31
Implement `postcard_schema::Schema` for `VecDeque`, `Rc`, and `Arc`

Calsign:
These are standard library types so I assume they are uncontroversial. I followed the existing conventions for these types.

(Side question: I understand the postcard-schema/postcard-schema-dyn split, but why do we need to implement the same schemas for both alloc and std, why can't std reuse the alloc implementations?)


tommasoclini:
is this going to be merged?

tommasoclini:
`portable-atomic-util`'s `Arc` could also be useful.

PR #291, opened 2026-02-03
[postcard-derive] Ignore zeroes in enum size calculation to avoid large if blocks

### LINKED ISSUE #691: `rustc` stack overflow when compiling rmk-types with Rust 1.92.0
After updating my stable toolchain to `rustc 1.92.0 (ded5c06cf 2025-12-08)` yesterday, I found that I could no longer compile any examples:

```
thread 'rustc' (6732) has overflowed its stack
error: could not compile `rmk-types` (lib)

```

The compilation only succeeds if I set the stack size to approximately 64MB via an environment variable: (pwsh)

```pwsh
$env:RUST_MIN_STACK = "67108864"
cargo build

```

or (bash):

```bash
export RUST_MIN_STACK=67108864
cargo build

```

I attempted to create a minimal reproduction and traced the issue to this specific location:
https://github.com/HaoboGu/rmk/blob/e7577ea3e28d7f54cd17f291af3f446bb71b8464/rmk-types/src/keycode.rs#L14-L17

The issue disappears if I either remove `#[derive(postcard::experimental::max_size::MaxSize)]` or reduce the overall size of the enum.

I suspect that Rust 1.92.0 might have introduced changes to macro expansion or other internals, as the compiler's default stack size remains unchanged:

https://github.com/rust-lang/rust/blob/123588a5bf9250f214e1931a761f814660e69620/compiler/rustc_interface/src/util.rs#L105

Could we perhaps file an issue with `postcard` or wait for the macro to stabilize, and maybe notify users that they might need to set `RUST_MIN_STACK` during compilation?

thealtofwar:
Resolves https://github.com/HaoboGu/rmk/issues/691.

This skips the zeroes in the size type calculation in order to compress the resulting tree generated.

### DIFF
--- a/source/postcard-derive/src/max_size.rs
+++ b/source/postcard-derive/src/max_size.rs
@@ -1,4 +1,4 @@
-use proc_macro2::{Span, TokenStream, Literal};
+use proc_macro2::{Span, TokenStream};
 use quote::{quote, quote_spanned};
 use syn::{
     parse_macro_input, parse_quote, spanned::Spanned, Data, DeriveInput, Fields, GenericParam,
--- a/source/postcard-derive/src/max_size.rs
+++ b/source/postcard-derive/src/max_size.rs
@@ -1,4 +1,4 @@
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::{Span, TokenStream, Literal};
 use quote::{quote, quote_spanned};
 use syn::{
     parse_macro_input, parse_quote, spanned::Spanned, Data, DeriveInput, Fields, GenericParam,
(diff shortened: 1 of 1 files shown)


source/postcard-derive/src/max_size.rs
```diff
@@ -41,11 +41,11 @@ fn add_trait_bounds(mut generics: Generics) -> Generics {
 /// Generate a constant expression that sums up the maximum size of the type.
 fn max_size_sum(data: &Data, span: Span) -> Result<TokenStream, syn::Error> {
     match data {
-        Data::Struct(data) => Ok(sum_fields(&data.fields)),
+        Data::Struct(data) => Ok(sum_fields(&data.fields).unwrap_or(quote!(0))),
         Data::Enum(data) => {
             let variant_count = data.variants.len();
 
-            let recurse = data.variants.iter().map(|v| sum_fields(&v.fields));
+            let recurse = data.variants.iter().map(|v| sum_fields(&v.fields)).filter_map(|v| v);
```
max-heller: Could this be `.iter().filter_map()` instead of `.iter().map().filter_map()`?

thealtofwar: Yes! Thanks for pointing it out.

max-heller (approved):
Looks good, thanks!

PR #299, opened 2026-04-30
[postcard-derive] Optimize derive(MaxSize) codegen

### LINKED ISSUE #286: [postcard-derive]: `MaxSize` codegen is too simple
https://github.com/HaoboGu/rmk/issues/691 is an issue that shows that when deriving `MaxSize` on an enum, if the enum has lots of variants, we generate a LOT of code and rustc runs out of stack.

We should improve this, even if the RMK issue is fixed by avoiding the derive.

max-heller:
Avoids generating nested expressions proportional to the number of variants, instead generating a slice of variant sizes and using a const function to determine the max size.

Also avoids generating unnecessary `0 + ` fragments.

Fixes #286

Before
```rust
impl ::postcard::experimental::max_size::MaxSize for Bar {
    const POSTCARD_MAX_SIZE: usize = 1usize
        + {
            let lhs = {
                let lhs = 0;
                let rhs = 0
                    + <u16 as ::postcard::experimental::max_size::MaxSize>::POSTCARD_MAX_SIZE;
                if lhs > rhs { lhs } else { rhs }
            };
            let rhs = 0
                + <u8 as ::postcard::experimental::max_size::MaxSize>::POSTCARD_MAX_SIZE;
            if lhs > rhs { lhs } else { rhs }
        };
}
```

After
```rust
impl ::postcard::experimental::max_size::MaxSize for Bar {
    const POSTCARD_MAX_SIZE: usize = 1usize
        + ::postcard::experimental::max_size::max_of_variants(
            &[
                <u16 as ::postcard::experimental::max_size::MaxSize>::POSTCARD_MAX_SIZE,
                <u8 as ::postcard::experimental::max_size::MaxSize>::POSTCARD_MAX_SIZE,
            ],
        );
}
```

### DIFF
--- a/source/postcard-derive-ng/src/max_size.rs
+++ b/source/postcard-derive-ng/src/max_size.rs
@@ -41,28 +41,17 @@ fn add_trait_bounds(mut generics: Generics) -> Generics {
 /// Generate a constant expression that sums up the maximum size of the type.
 fn max_size_sum(data: &Data, span: Span) -> Result<TokenStream, syn::Error> {
     match data {
-        Data::Struct(data) => Ok(sum_fields(&data.fields)),
+        Data::Struct(data) => Ok(sum_fields(&data.fields).unwrap_or(quote!(0))),
         Data::Enum(data) => {
             let variant_count = data.variants.len();
 
-            let recurse = data.variants.iter().map(|v| sum_fields(&v.fields));
+            let variant_sizes = data.variants.iter().filter_map(|v| sum_fields(&v.fields));
 
             let discriminant_size = varint_size_discriminant(variant_count as u32) as usize;
 
-            // Generate a tree of max expressions.
-            let max = recurse.fold(quote!(0), |acc, x| {
-                quote! {
-                    {
-                        let lhs = #acc;
-                        let rhs = #x;
-                        if lhs > rhs {
-                            lhs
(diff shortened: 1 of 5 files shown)


source/postcard/Cargo.toml
```diff
@@ -69,6 +69,7 @@ optional = true
 [dependencies.postcard-derive]
 version = "0.2.0"
 optional = true
+path = "../postcard-derive"
```
max-heller, line 72: We weren't actually using the local copy of `postcard-derive` in tests, this change mirrors what we do for all other workspace dependencies

source/postcard-derive-ng/src/max_size.rs
```diff
@@ -41,28 +41,17 @@ fn add_trait_bounds(mut generics: Generics) -> Generics {
...
-                        if lhs > rhs {
-                            lhs
-                        } else {
-                            rhs
-                        }
-                    }
-                }
-            });
+            let max = quote! {
+                ::postcard2::experimental::max_size::max_of_variants(&[#(#variant_sizes),*])
```
max-heller, line 53: @jamesmunns `postcard-derive-ng` already references `postcard2::experimental::max_size` below, but that module doesn't seem to exist in `postcard2`

PR #300, opened 2026-07-11
Implement MaxSize and Schema for enum_map::EnumMap

sugar700:
This was requested in `enum_map`, with the pull request at https://codeberg.org/sugar700/enum-map/pulls/170, however I decided against that due to experimental status of `MaxSize` trait.

I also implemented `Schema`, mostly because I could do so.

### DIFF
--- a/source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
+++ b/source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
@@ -1,0 +1,9 @@
+//! Implementations of the [`Schema`] trait for the `enum-map` crate v3
+
+use crate::{schema::DataModelType, Schema};
+
+#[cfg_attr(docsrs, doc(cfg(feature = "enum-map-v3_0")))]
+impl<K: enum_map_v3_0::Enum<Array<V>: Schema>, V: Schema> Schema for enum_map_v3_0::EnumMap<K, V> {
+    // K::Array is guaranteed to be [T; N]
+    const SCHEMA: &'static DataModelType = K::Array::SCHEMA;
+}
--- a/source/postcard-schema-ng/Cargo.toml
+++ b/source/postcard-schema-ng/Cargo.toml
@@ -85,6 +85,12 @@ package = "serde-big-array"
 version = "0.5.1"
 optional = true
 
+[dependencies.enum-map_v3_0]
+package = "enum-map"
+version = "3.0.0"
+default-features = false
+optional = true
+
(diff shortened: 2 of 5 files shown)


sugar700:
May as well close this, I released enum-map 3.0.0, and currently it's not exactly possible to implement those traits for the newest version due to changes I did to make the interface nicer, due to missing language features.

sugar700:
Or, actually, no, it is possible.

sugar700:
Updated the pull request to be for enum-map 3.0.0.

source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
```diff
@@ -0,0 +1,8 @@
+//! Implementations of the [`Schema`] trait for the `enum-map` crate v2
```
max-heller:
```suggestion
//! Implementations of the [`Schema`] trait for the `enum-map` crate v3
```

source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
```diff
@@ -0,0 +1,8 @@
...
+
+use crate::{schema::DataModelType, Schema};
+
+#[cfg_attr(docsrs, doc(cfg(feature = "enum-map-v3_0")))]
+impl<K: enum_map_v3_0::Enum, V: Schema> Schema for enum_map_v3_0::EnumMap<K, V> {
+    const SCHEMA: &'static DataModelType = &DataModelType::Seq(V::SCHEMA);
```
max-heller: The `Serialize` implementation for `EnumMap` uses `serialize_tuple()`, not `serialize_seq()`, so this should be `::Tuple(...)`. The two have different wire layouts: sequences are length-prefixed variable-length collections whereas tuples are fixed-length concatenations of their contents

max-heller (changes requested):

source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
sugar700: Right, that makes sense. Fixed.

source/postcard-schema-ng/src/impls/enum_map_v3_0.rs
sugar700: Fixed.

max-heller (approved):

max-heller:
Thanks for the contribution!

Issue #217, opened 2025-02-11
[postcard-schema] Key hash calculation could suffer collision if path and schema alias


jamesmunns:
Described in https://github.com/jamesmunns/postcard-spec-ng/issues/1#issue-2845133457,

Since some of the schema primes fall in the ascii range (and really, many other values could fall in the utf-8 range), it could be possible to get a collision between two Keys, for example since `b'e'` is the prime used for `bytearray`, the following two would alias:

```
("car", (bytearray, bytearray))
("care", (bytearray))
```

We should consider this to the next breaking change in `postcard-schema` and `postcard-rpc`.

Right now in `postcard-rpc`, this will lead to a compilation failure, as it would detect the two keys collide.

CC @DavidBuchanan314 who discovered this, and @max-heller for thoughts.

jamesmunns:
One proposed fix for this is adding a value to be hashed between the two, ideally something that is NEITHER valid utf-8 NOR a valid schema item. We could potentially hash a fragment that is invalid utf-8, and also not a valid schema prime identifier.

DavidBuchanan314:
Another mitigation could be to length-prefix the path

max-heller:
Length-prefixing seems like a good solution.

> we accept an empty path string, which probably isn't great for some reason I can't think of right now.

I'll note that I use empty paths

jamesmunns:
One thing to note about length prefixing: we might need to pick a max length (e.g. `u16`, `u32`, `u64`), or commit to using varint encoding (`varint(usize)`) for the length. Basically, this is a note to not just blindly use `usize` as the length and hash the `to_le_bytes` form.

thaodt:
> One thing to note about length prefixing: we might need to pick a max length (e.g. `u16`, `u32`, `u64`), or commit to using varint encoding (`varint(usize)`) for the length. Basically, this is a note to not just blindly use `usize` as the length and hash the `to_le_bytes` form.

I'd like to go with using varint encoding for length prefixing. Fixed length can waste space for small values (e.g., u32 uses 4 bytes to store "testing" despite its 7-byte length).

jamesmunns:
[...]

Issue #192, opened 2024-11-28
Additional Metadata for Schema?


jamesmunns:
CC @max-heller and https://github.com/jamesmunns/postcard/pull/179

There have been a couple of asks for additional schema metadata. Off the top of my head:

* Things like max size (for bounded types, and possibly annotations for unbounded types)
* Things like descriptions/comments (though some of this veers into `postcard-rpc`'s `Endpoint`s and `Topic`s, which tend to benefit from metadata as well

Open questions would be:

* Should (some or all of) these fields affect the schema hash calculation?
* How can users opt-in/out of sending this information over the wire?

max-heller:
- Min size: #180
- Custom enum discriminants?

jamesmunns:
Whatcha mean by custom enum discriminants? (Postcard specifically states that it uses "lexical ordering")

max-heller:
> Whatcha mean by custom enum discriminants? (Postcard specifically states that it uses "lexical ordering")

```rust
enum Foo {
    A = 1,
    ...
}
```
Similar to comments, serde and postcard don't care about discriminants, but some use cases of `Schema` as a reflection mechanism might.

max-heller:
> Things like max size (for bounded types, and possibly annotations for unbounded types)

Annotations as in something like this?
```rust
#[postcard(serialized_size(max = 512))]
bytes: Vec<u8>,
```
 
> Should (some or all of) these fields affect the schema hash calculation?

One way this could work is by having a wrapper type for customizing hashing behavior:

```rust
struct HashBy<T> {
    // Which fields to include in the hash
    fields: Fields,
    value: T,
}
// Could be a bitset or something more compact
struct Fields {
    names: bool,
    max_size: bool,
    ...
}
impl Hash for HashBy<NamedType> {}
...
```

> How can users opt-in/out of sending this information over the wire?

I could see this working with a `SerializeWith<T>` wrapper (similar to the one above for hashing) combined with optional/defaulted fields for max size, comments, etc. on the deserializing side.

One other open question: 
- How much will additional (unused) metadata affect type and binary sizes? Ideally it could be optimized out and comment strings wouldn't end up getting embedded in binaries if only the basic schema is needed, but I'm not sure just how smart the compiler is with `const`s

jamesmunns:
re: hashing, I specifically meant what `postcard-rpc` does for creating a `Key` from a `NamedType`.

[...]

Issue #180, opened 2024-11-08
Feature request: MinSize


elpiel:
Apart from the MaxSize experimental feature I'd love to see a MinSize as well.
This can be useful if you want to have a "counter" buffer which keeps track of the maximum number of items that might be able to fit in a given buffer.

E.g. I have a buffer of 4 KiB and I'd like to know what's the maximum number of messages that can fit inside.
This buffer contains `embedded_sdmmc::Block`s inside so it's not a continues buffer of bytes and I'm using a second buffer to keep track of how much each block is filled.

jamesmunns:
I somewhat plan to rework MaxSize as part of `postcard-schema`, as I believe it is possible to calculate this as a const fn. See https://github.com/jamesmunns/postcard/pull/179 for an example of this.

One edge case I can think of is how to handle "infinite" cases, as `()` is serialized as zero bytes on the wire, so your function needs to be careful not to divide by zero :)

jamesmunns:
If you'd like to copy and paste what I did in #179 and switch it to min instead of max, I'd accept that PR.

elpiel:
Hmm that's an interesting case actually heh.
Well I do have a "protocol" level crate although I only (currently) use it with 1 enum that hosts all the messages and it doesn't have a unit type.

PR #306, opened 2026-09-05
postcard-schema-ng: Implement bounding on types

jamesmunns:
This allows us to calculate max size.

This is a second attempt at what I started in #179.

### DIFF
--- a/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
+++ b/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
@@ -4,5 +4,56 @@ use crate::{schema::DataModelType, Schema};
 
 #[cfg_attr(docsrs, doc(cfg(feature = "chrono-v0_4")))]
 impl<Tz: chrono_v0_4::TimeZone> Schema for chrono_v0_4::DateTime<Tz> {
-    const SCHEMA: &'static DataModelType = &DataModelType::String { bounds: Some(32) };
+    // Chrono serializes as an rfc3339 string, which is typically about 27 bytes:
+    //
+    // "2026-09-06T13:24:04.659820Z"
+    //  000000000011111111112222222
+    //  012345678901234567890123456
+    //
+    // But you can also make it spit out somewhat silly output with custom precision
+    // and time zones, see `longest_rfc3339` below. If you can make chrono output a
+    // longer serialized output, please open an issue.
+    //
+    // "+262142-12-31T23:59:59.999999999+23:59"
+    //  00000000001111111111222222222233333333
+    //  01234567890123456789012345678901234567
+    const SCHEMA: &'static DataModelType = &DataModelType::String { bounds: Some(38) };
+}
+
+#[cfg(test)]
+mod test {
+    use crate::Schema;
+    use chrono_v0_4::{DateTime, FixedOffset, NaiveDate, TimeZone, Utc};
+
+    fn longest_rfc3339() -> DateTime<FixedOffset> {
+        let offset = FixedOffset::east_opt(23 * 3600 + 59 * 60).unwrap();
+        let local = NaiveDate::MAX
+            .and_hms_nano_opt(23, 59, 59, 999_999_999)
+            .unwrap();
+        offset.from_local_datetime(&local).single().unwrap()
+    }
+
+    #[test]
+    fn chrono_longest_check() {
+        let dt = longest_rfc3339();
+        let mut buf = [0u8; 64];
+        let used = postcard::to_slice(&dt, &mut buf).unwrap();
+        assert_eq!(
+            used.len(),
+            DateTime::<FixedOffset>::SCHEMA.max_size().unwrap()
+        );
+        assert_eq!(
+            postcard::from_bytes::<DateTime<FixedOffset>>(used).unwrap(),
+            dt
+        );
+    }
+
+    #[test]
+    fn chrono_basic_check() {
+        let dt = Utc::now();
+        let mut buf = [0u8; 64];
+        let used = postcard::to_slice(&dt, &mut buf).unwrap();
+        assert!(used.len() <= DateTime::<FixedOffset>::SCHEMA.max_size().unwrap());
+        assert_eq!(postcard::from_bytes::<DateTime<Utc>>(used).unwrap(), dt);
+    }
 }
--- a/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
+++ b/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
@@ -4,5 +4,5 @@ use crate::{schema::DataModelType, Schema};
 
 #[cfg_attr(docsrs, doc(cfg(feature = "chrono-v0_4")))]
 impl<Tz: chrono_v0_4::TimeZone> Schema for chrono_v0_4::DateTime<Tz> {
-    const SCHEMA: &'static DataModelType = &DataModelType::String;
+    const SCHEMA: &'static DataModelType = &DataModelType::String { bounds: Some(32) };
 }
--- a/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
+++ b/source/postcard-schema-ng/src/impls/chrono_v0_4.rs
@@ -17,7 +17,7 @@ impl<Tz: chrono_v0_4::TimeZone> Schema for chrono_v0_4::DateTime<Tz> {
     // "+262142-12-31T23:59:59.999999999+23:59"
     //  00000000001111111111222222222233333333
     //  01234567890123456789012345678901234567
-    const SCHEMA: &'static DataModelType = &DataModelType::String { bounds: Some(38) };
+    const SCHEMA: &'static DataModelType = &DataModelType::String { max_len: Some(38) };
 }
 
 #[cfg(test)]
--- a/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
+++ b/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
@@ -4,8 +4,8 @@ use crate::{schema::DataModelType, Schema};
 
 impl Schema for uuid_v1_0::Uuid {
     const SCHEMA: &'static DataModelType = &DataModelType::Seq {
-        item: &DataModelType::U8,
-        bounds: Some(16),
+        element: &DataModelType::U8,
+        max_len: Some(16),
     };
 }
 
--- a/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
+++ b/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
@@ -3,5 +3,20 @@
 use crate::{schema::DataModelType, Schema};
 
 impl Schema for uuid_v1_0::Uuid {
-    const SCHEMA: &'static DataModelType = &DataModelType::ByteArray { bounds: Some(16) };
+    const SCHEMA: &'static DataModelType = &DataModelType::Seq {
+        item: &DataModelType::U8,
+        bounds: Some(16),
+    };
+}
+
+#[cfg(test)]
+mod test {
+    use crate::Schema;
+    #[test]
+    fn basic() {
+        let uuid = uuid_v1_0::uuid!("21E764AB-6B44-42D3-BE73-682E5E85C196");
+        let mut buf = [0u8; 17];
+        let res = postcard::to_slice(&uuid, &mut buf).unwrap();
+        assert_eq!(res.len(), uuid_v1_0::Uuid::SCHEMA.max_size().unwrap());
+    }
 }
--- a/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
+++ b/source/postcard-schema-ng/src/impls/uuid_v1_0.rs
@@ -3,5 +3,5 @@
 use crate::{schema::DataModelType, Schema};
 
 impl Schema for uuid_v1_0::Uuid {
-    const SCHEMA: &'static DataModelType = &DataModelType::ByteArray;
+    const SCHEMA: &'static DataModelType = &DataModelType::ByteArray { bounds: Some(16) };
 }
--- a/source/postcard-schema-ng/src/key/hash.rs
+++ b/source/postcard-schema-ng/src/key/hash.rs
@@ -65,6 +65,24 @@ pub mod fnv1a64 {
         hash_sdm_type(state, schema).to_le_bytes()
     }
 
+    pub(crate) const fn hash_bounds(state: u64, bounds: Option<usize>) -> u64 {
+        let mut state = hash_update(state, &[0x2B]);
+        let Some(bound) = bounds else {
+            return state;
+        };
+        let mut value = bound;
+        while value != 0 {
+            let byte = value.to_le_bytes()[0];
+            if value < 128 {
+                return hash_update(state, &[byte]);
+            }
+
+            state = hash_update(state, &[byte | 0x80]);
+            value >>= 7;
+        }
+        state
+    }
+
     pub(crate) const fn hash_update(mut state: u64, bytes: &[u8]) -> u64 {
         let mut idx = 0;
         while idx < bytes.len() {
@@ -110,6 +128,8 @@ pub mod fnv1a64 {
         //     0x1F, 0xA3, 0x35, 0x43, 0x89, 0x49, 0xE3, 0x07,
         //     0x53, 0xF1, 0x17, 0x2F, 0x29, 0x59,
         // ];
+        //
+        // 0x2B reserved for bounds
         match sdmty {
             DataModelType::Bool => hash_update(state, &[0x11]),
             DataModelType::I8 => hash_update(state, &[0xC5]),
@@ -127,16 +147,23 @@ pub mod fnv1a64 {
             DataModelType::F32 => hash_update(state, &[0xEF]),
             DataModelType::F64 => hash_update(state, &[0x71]),
             DataModelType::Char => hash_update(state, &[0xC1]),
-            DataModelType::String => hash_update(state, &[0x25]),
-            DataModelType::ByteArray => hash_update(state, &[0x65]),
+            DataModelType::String { bounds } => {
+                let state = hash_update(state, &[0x25]);
+                hash_bounds(state, *bounds)
+            }
+            DataModelType::ByteArray { bounds } => {
+                let state = hash_update(state, &[0x65]);
+                hash_bounds(state, *bounds)
+            }
             DataModelType::Option(t) => {
                 let state = hash_update(state, &[0x6D]);
                 hash_sdm_type(state, t)
             }
             DataModelType::Unit => hash_update(state, &[0x47]),
-            DataModelType::Seq(t) => {
+            DataModelType::Seq { item: t, bounds } => {
                 let state = hash_update(state, &[0x03]);
-                hash_sdm_type(state, t)
+                let state = hash_sdm_type(state, t);
+                hash_bounds(state, *bounds)
             }
             DataModelType::Tuple(ts) => {
                 let mut state = hash_update(state, &[0xA7]);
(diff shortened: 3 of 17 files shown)


source/postcard-schema-ng/src/key/hash.rs
```diff
@@ -431,7 +469,7 @@ mod test {
 
         assert_eq!(
             hash_ty_path::<Bar>("test_path"),
-            [139, 128, 52, 27, 107, 8, 218, 98]
+            [224, 143, 54, 58, 255, 237, 252, 44]
```
jamesmunns: I don't know if I love this. I could leave `bounds: None` as not doing an additional hashing step, which would keep unbounded collections the same, but it would still be a breaking hash change for collections like `heapless::Vec<T, N>`.

source/postcard-schema-ng/src/impls/chrono_v0_4.rs
```diff
@@ -4,5 +4,5 @@ use crate::{schema::DataModelType, Schema};
 
 #[cfg_attr(docsrs, doc(cfg(feature = "chrono-v0_4")))]
 impl<Tz: chrono_v0_4::TimeZone> Schema for chrono_v0_4::DateTime<Tz> {
-    const SCHEMA: &'static DataModelType = &DataModelType::String;
+    const SCHEMA: &'static DataModelType = &DataModelType::String { bounds: Some(32) };
```
max-heller: How do we derive this bound? Would be helpful to add a comment

source/postcard-schema-ng/src/schema/mod.rs
```diff
@@ -84,10 +81,16 @@ pub enum DataModelType {
     Char,
 
     /// The `String` Serde Data Model Type
-    String,
+    String {
+        /// Upper bound of items in the sequence
+        bounds: Option<usize>,
```
max-heller: For all types: `bounds` to me implies multiple bounds (upper/lower); if we're only supporting an upper bound (which I think is reasonable, lower bounds do not seem nearly as useful except for things like calculating `min_size()` (https://github.com/jamesmunns/postcard/issues/180)), then I'd lean towards a name like `max_len` or `upper_bound` or something else that's more explicit.

For strings in particular: would be good to clarify that this is the number of bytes (similar to the docs of `str::len()`)

source/postcard-schema-ng/src/schema/mod.rs
```diff
@@ -129,6 +139,120 @@ pub enum DataModelType {
...
+                    I128 => Some(postcard_core::varint::varint_max::<i128>()),
+                    U16 => Some(postcard_core::varint::varint_max::<u16>()),
+                    U32 => Some(postcard_core::varint::varint_max::<i32>()),
+                    U64 => Some(postcard_core::varint::varint_max::<u64>()),
+                    U128 => Some(postcard_core::varint::varint_max::<u128>()),
+                    Usize => Some(postcard_core::varint::varint_max::<usize>()),
+                    Isize => Some(postcard_core::varint::varint_max::<isize>()),
+                    F32 => Some(4),
+                    F64 => Some(8),
+                    Char => Some(5),
```
max-heller, line 187: This comes from serializing a slice of up to four bytes, right? So one length byte plus up to four data bytes?

source/postcard-schema-ng/src/schema/mod.rs
```diff
@@ -129,6 +139,120 @@ pub enum DataModelType {
...
+                        };
+                        let pair = ksize + vsize;
+                        let pairs = (*bound) * pair;
+                        Some(size_as_varint_usize(*bound) + pairs)
+                    }
+                    Struct { name: _, data } => data.max_size(),
+                    Enum { name: _, variants } => {
+                        let mut idx = 0;
+                        let disc = if variants.is_empty() {
+                            1
```
max-heller, line 232: If there are no postcard-serializable variants, shouldn't it be impossible to serialize this enum, and therefore have a max size of 0? 1 seems like a fine fallback as well for this impossible case, just curious

source/postcard-schema-ng/src/schema/mod.rs
```diff
@@ -129,6 +139,120 @@ pub enum DataModelType {
...
+                                return None;
+                            };
+                            if size > max {
+                                max = size;
+                            }
+                            idx += 1;
+                        }
+                        Some(disc + max)
+                    }
+                    Schema => todo!(),
```
max-heller: Does `Schema` have a max size? Should this be `None`?

source/postcard-schema-ng/src/schema/mod.rs
```diff
@@ -96,7 +99,12 @@ pub enum DataModelType {
     Unit,
 
     /// The "Sequence" Serde Data Model Type
-    Seq(&'static Self),
+    Seq {
+        /// Items in the sequence
+        item: &'static Self,
```
max-heller: [`serde` uses `element` to refer to items in a sequence](https://docs.rs/serde/latest/serde/ser/trait.SerializeSeq.html#tymethod.serialize_element), would be nice to be consistent

source/postcard-schema-ng/src/schema/fmt.rs
```diff
@@ -33,13 +33,16 @@ pub fn is_prim(osdmty: &OwnedDataModelType) -> bool {
         OwnedDataModelType::F32 => true,
         OwnedDataModelType::F64 => true,
         OwnedDataModelType::Char => true,
-        OwnedDataModelType::String => true,
-        OwnedDataModelType::ByteArray => true,
+        // TODO(AJM) is this right?
+        OwnedDataModelType::String { bounds } => bounds.is_none(),
+        OwnedDataModelType::ByteArray { bounds } => bounds.is_none(),
```
max-heller: What do we want `is_prim()` to guarantee? It's impossible for it to guarantee that the type whose schema we're looking at is literally `str` or `[u8]` (vs something like `String` or `Vec<u8>`), so I'd lean towards ignoring bounds here as `heapless::String<10>` feels comparable to `String`)

source/postcard-schema-ng/src/schema/fmt.rs
```diff
@@ -105,18 +108,31 @@ pub fn fmt_owned_dmt_to_buf(dmt: &OwnedDataModelType, buf: &mut String, top_leve
         OwnedDataModelType::F32 => *buf += "f32",
         OwnedDataModelType::F64 => *buf += "f64",
         OwnedDataModelType::Char => *buf += "char",
-        OwnedDataModelType::String => *buf += "String",
-        OwnedDataModelType::ByteArray => *buf += "[u8]",
+        OwnedDataModelType::String { bounds } => {
+            *buf += "String";
```
max-heller, line 114: If we're doing a breaking release already, would be nice to change this to `str`

source/postcard-schema-ng/src/schema/fmt.rs
```diff
@@ -105,18 +108,31 @@ pub fn fmt_owned_dmt_to_buf(dmt: &OwnedDataModelType, buf: &mut String, top_leve
...
+            if let Some(bound) = bounds {
+                *buf += &format!("[..{bound}]");
+            }
+        }
+        OwnedDataModelType::ByteArray { bounds } => {
+            *buf += "[u8]";
+            if let Some(bound) = bounds {
+                *buf += &format!("[..{bound}]");
```
max-heller, line 122: I don't have a strong preference either way, but we could consider something like `[u8; ..{bound}]` to be more concise. This would also work for sequences, but not for strings or maps, so maybe this uniform suffix is better

source/postcard-schema-ng/src/key/hash.rs
max-heller: My preference here would be to hash only the pieces that break wire compatibility if changed (i.e. only the serde-relevant pieces), at least by default

Prior discussion: https://github.com/jamesmunns/postcard/issues/192#issuecomment-2506726286

max-heller:
Thanks for reviving this! Very excited to remove our `MaxSize` derive

source/postcard-schema-ng/src/impls/chrono_v0_4.rs
jamesmunns: Ah, I should. I looked at chrono's rfc3339 output, and counted the output in the worst case (with nanos precision), which IIRC was 29 or 30, and I added a fudge factor. I should actually add a test case here for this though and not round up.

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: I think that's reasonable, it is definitely just an "upper bounds".

I feel like "MinSize" would be basically the same as the `fn max_size` impl with all of the current upper bounds treated as zero?.

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: Yep, first byte can be 1..=4, then the 1..=4 bytes of the char. (this is a terrible encoding form, there's an issue to fix this if we ever do a breaking wire change).

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: That's a good point. I might try to add a test around this, but an enum with zero variants does seem a bit silly, wire wise.

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: This should probably be `None`, yeah.

source/postcard-schema-ng/src/schema/mod.rs
max-heller: Leaning towards `max_len` over `upper_bound` so the same name makes sense for strings as well as other collections

max-heller: > I feel like "MinSize" would be basically the same as the fn max_size impl with all of the current upper bounds treated as zero?.

It can be, though in the presence of types like `NonEmptyVec` or similar it would be an underestimate

source/postcard-schema-ng/src/schema/fmt.rs
jamesmunns: I need to look again what I was using this for. I think I used it for postcard-rpc or poststation to avoid sending "builtin" schemas, this may be different now in postcard-schema-ng where the shape of schemas have changed and we don't have top level types like `u8` containing a named data type.

I may just remove it for now, we can add it back later if we need.

source/postcard-schema-ng/src/schema/fmt.rs
jamesmunns: oooh, that's a better syntax, and yeah, I struggled with what this should be.

source/postcard-schema-ng/src/key/hash.rs
jamesmunns: Yeah, let me read that back. At least from the perspective of postcard-rpc, I am somewhat tempted to say "if you change the max size, that's a wire compatibility change".

I know this is sort of a grey area though, if you HAPPEN to have enough space it totally will decode right (and therefore "the wire representations are the same"), but from the perspective of "do both sides agree on the same schema", I think different sizes should be treated as different hashes.

max-heller: At $WORK we depend on being able to change the max length in a backwards-compatible way in some places (we have a dedicated `TruncatingVec` type that will accept as many elements as it can fit). Going back to the '24 discussion, we could expose options for how to calculate the hash, and allow users to decide what they want? If postcard-rpc wants to take the strict interpretation that seems like a sane answer, but I don't think we should force it on all users

jamesmunns: Yeah, I think the right way to handle this is to change the hasher to include a `config` item (in addition to the current u64 state), and allow the caller to pass in that config which decides whether bounds (or future metadata) is considered as part of the hashing. I haven't done that yet, but will.

source/postcard-schema-ng/src/impls/chrono_v0_4.rs
jamesmunns: Added a test for this, 32 was underconsidering it. I think I have an upper bound now, but it's potentially questionable whether we consider this bounded or not.

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: Renamed to max_len, and added `MaxLen*` types that include these bounds.

Good call on the NonEmpty items. Maybe some day we want to add `min_len` as metadata, I don't totally understand the desire for it though practically, so maybe I'll leave it for a future PR. I wonder if we should make some more things `non_exhaustive` to avoid this being a breaking change.

source/postcard-schema-ng/src/schema/mod.rs
jamesmunns: Fixed.

source/postcard-schema-ng/src/schema/fmt.rs
jamesmunns: By the way, why "str" and not "String"? I feel like either are equally valid.

jamesmunns:
I do still need to update postcard-dyn.

source/postcard-schema-ng/src/key/hash.rs
jamesmunns: @max-heller one thing that I realized when thinking about this: if you're doing what you describe, and you start using the `max_size()` stuff to auto-size buffers, won't that end up biting you if someone sends an array that's too big, and that pushes the size of a single message over the calculated `max_size()`?

This may not be a problem if you're using something else as the max length, like if you say a frame can't exceed an MTU of 1544 bytes or something, which ends up being the forcing function instead. But if you you're buffering serial data for example, and with a `TruncatingVec<24>` you calculate your max RX size as 300 bytes, but someone sends a `TruncatingVec<32>` and ends up sending you a 400 byte message, you'll overflow your incoming buffer entirely and still discard the message.

I still think "add a config to the hasher" is probably the right take, but it does seem *somewhat* foot-gunny to use auto-sized buffers using `max_size` *without* enabling the hashing of bounds.

source/postcard-schema-ng/src/impls/uuid_v1_0.rs
```diff
@@ -3,5 +3,20 @@
 use crate::{schema::DataModelType, Schema};
 
 impl Schema for uuid_v1_0::Uuid {
-    const SCHEMA: &'static DataModelType = &DataModelType::ByteArray { bounds: Some(16) };
+    const SCHEMA: &'static DataModelType = &DataModelType::Seq {
```
max-heller, line 6: I think `ByteArray` was correct given that the implementation uses `serialize_bytes()`: https://docs.rs/uuid/1.26.0/src/uuid/external/serde_support.rs.html#31

source/postcard-schema-ng/src/schema/fmt.rs
max-heller: `str` is to `String` what `[u8]` (which we use for `ByteArray`) is to `Vec<u8>` (or `Box<[u8]>`), which is to say the underlying primitive type -- and serde data model type

source/postcard-schema-ng/src/impls/chrono_v0_4.rs
max-heller: I think it's fine either way. I'd lean towards avoiding the bound since RFC3339 seems to technically allow unbounded length

source/postcard-schema-ng/src/schema/mod.rs
max-heller: > so maybe I'll leave it for a future PR

Agreed this makes sense for a separate PR if at all

> I wonder if we should make some more things non_exhaustive to avoid this being a breaking change.

This could make sense, though we should avoid making it a pain to manually implement `Schema`. One option would be to take inspiration from [`syn` 3.0](https://github.com/dtolnay/syn/releases/tag/3.0.0) (see "Modifiers") and add `non_exhaustive` structs as fields in places we expect to want to add more data in the future, see e.g. https://docs.rs/syn/3.0.5/syn/struct.Field.html#structfield.modifiers

source/postcard-schema-ng/src/key/hash.rs
```diff
@@ -65,6 +65,24 @@ pub mod fnv1a64 {
         hash_sdm_type(state, schema).to_le_bytes()
     }
 
+    pub(crate) const fn hash_bounds(state: u64, bounds: Option<usize>) -> u64 {
+        let Some(bound) = bounds else {
+            return state;
+        };
+        let mut state = hash_update(state, &[0x2B]);
```
max-heller, line 72: When we are going to be hashing bounds (according to the hashing options), don't we need to hash `0x2B` everywhere the optional bounds appear (even if `None`) to avoid hash collisions? If we only hash this prime when bounds *are* present, I think we can get into ambiguities where something like `{ name: [0x1], max_len: Some(0x5) }` feeds the same bytes into the hasher as `{ name: [0x1, 0x2B, 0x5], max_len: None }`

source/postcard-schema-ng/src/key/hash.rs
max-heller: That's a good point, that can certainly bite you. On the one hand, that outcome is bad, so it's worth discouraging; on the other, if we require hashing the max length, then the message would have been rejected anyway because of a hash mismatch, so you aren't much better off.

In practice, I think for things like this we usually have a fixed buffer size (MTU/max IP packet size/etc.) or a max size that far exceeds the expected size of any given packet because of many `None`s inside.

The other use-case that I find valuable that depends on ignoring max length in hashing is sending `&[T]` where `heapless::Vec<T, N>` is expected (and other equivalent pairs). This avoids needing to stack-allocate a `Vec` just to serialize it again. This could probably be solved via bounded, borrowed slice types like the ones you've added, but that may not work in every case.

source/postcard-schema-ng/src/impls/uuid_v1_0.rs
jamesmunns: Yeah... postcard doesn't actually discriminate well between "byte array" and "[u8]", and we declare the schema of `[u8]` as

```
impl<T: Schema> Schema for [T] {
    const SCHEMA: &'static DataModelType = &DataModelType::Seq {
        element: T::SCHEMA,
        max_len: None,
    };
}
```

I wonder if we should actually just discard `ByteArray` in `-ng` to avoid this being confusing.

source/postcard-schema-ng/src/key/hash.rs
jamesmunns: I think https://github.com/jamesmunns/postcard/issues/217 captures this, we should probably fix that for -ng.

source/postcard-schema-ng/src/impls/uuid_v1_0.rs
max-heller: Given [it's a distinct type in the data model](https://serde.rs/data-model.html#types), I think it's important to keep `ByteArray` around -- serializers can choose to serialize them differently than a sequence of bytes (though Postcard doesn't)

max-heller: The fact that `[u8]` can't be specialized to use `serialize_bytes()` is more of an implementation limitation than anything

source/postcard-schema-ng/src/key/hash.rs
max-heller: That probably covers it... though it still gives me a bad feeling to do it conditionally

source/postcard-schema-ng/src/impls/uuid_v1_0.rs
jamesmunns: > serializers can choose to serialize them differently than a sequence of bytes (though Postcard doesn't)

Are you using `postcard-schema` for not-postcard-format things? A long while back you asked "is postcard-schema for postcard only, or serde in general", and these days, I'm feeling a bit stronger about it only being for postcard.

max-heller: > Are you using postcard-schema for not-postcard-format things?

Kinda sorta not really. `postcard-schema` is *close* to being useful as a general `serde` schema, but I've somewhat lost hope in full fledged dynamic (de)serialization with serde. Where possible though (including here), I think it's still valuable to have a 1:1 match between `postcard-schema` and `serde`'s data model or other abstractions.

